PR #1, opened 2016-07-30
Rename template_files/ to service/ in t-gokit for symmetry

hasAdamr:


hasLeland:
Looks like a great change, makes things easier!

PR #2, opened 2016-07-30
Truss go test

hasAdamr:


hasLeland:
despite the misgivings raised by @hasAdamr, I'm not altogether worried about the "ugliness" of this testing code. It's relatively straightforward code, and I'm pretty sure there's not a more idiomatic way of getting this code to do what we want.

Since what we want is to run large "is everything working" tests and not small clean unit tests, and we want theses tests to be runnable via the `go test` tooling, it seems that this approach is the best fit, and it's fundamental hackish nature is a bit unavoidable. For this reason, I'm fine with merge this in as is.

PR #4, opened 2016-08-01
Put back cleantest.sh to allow manual removal of generated test files

hasAdamr:
In using the integration tests I noticed I needed to still manually remove test files sometimes. I have added cleantest.sh back so that the Makefile can remove them.


hasLeland:
It's possible to pass a custom flag to a test, which you could use to clean out the old files: https://golang.org/cmd/go/#hdr-Description_of_testing_flags

I think that's a better alternative than bringing back in a shell file. Thoughts?

PR #7, opened 2016-08-01
Remove and exclude breaking vendored files

hasLeland:
Take two, this time hopefully not breaking the build. You'll probably want to `rm -rf` your `vendor/` before pulling these changes in, to make sure you don't have the problem files included.


hasAdamr:
Actually tried it this time! Works!

\ʕ◔ϖ◔ʔ/

PR #8, opened 2016-08-02
Refactor truss and integration_tests

hasAdamr:
truss has been refactored into main.go and generator/generator.go.
Where main.go deals with working directory, GOPATH, and argument
parsing and sanitation and generator.go actually exec's protoc and go
build to generate and build the microservice.

integration_tests were refactored to use `testing.T` to do logging
rather than logrus. As well as be more readable. It can also be called
with `$ go test -tags=integration -clean` in order to remove the
generated test files and nothing else.

Eventual goals are to unit test generator.go's functions and
eventually move the integration_tests to the generator package and test
by calling functions rather than execing the truss binary


hasLeland:
I note that the integration tests doesn't give any feedback on whats running, so it just sits there and prints no output till the integration tests all complete, even when invoked with the `-v` option. Given that good output was a big focus of the prior integration runner, I think that should be in the work that's ultimately merged.

---

When attempting to use the `make test`, it fails due to being in the wrong directory. It's attempting to run from the directory `truss/` but for the `go test` command to work we have to be in the `truss/integration_tests/`.

To have `make` continue working, the `test` rule should have a command of `go test ./... -tags=integration`.

However, if I'm in the `truss/` directory, and I run `go test ./... -tags=integration -clean` I get lots of errors (though all the files are correctly deleted) which look like this:

```
?       github.com/TuneLab/gob/truss    [no test files]
?       github.com/TuneLab/gob/truss/generator  [no test files]
go build github.com/TuneLab/gob/truss/integration_tests/test_service_definitions/add/service/DONOTEDIT/third_party/googleapis/google/api: /home/leland/bin/go/pkg/tool/linux_amd64/compile: chdir /home/leland/projects/go-projects/src/github.com/TuneLab/gob/truss/integration_tests/test_service_definitions/add/service/DONOTEDIT/third_party/googleapis/google/api: no such file or directory
go build github.com/TuneLab/gob/truss/integration_tests/test_service_definitions/add/service/DONOTEDIT/client/http: /home/leland/bin/go/pkg/tool/linux_amd64/compile: chdir /home/leland/projects/go-projects/src/github.com/TuneLab/gob/truss/integration_tests/test_service_definitions/add/service/DONOTEDIT/client/http: no such file or directory
```

I don't know if we care about those errors, as I don't get them when I run the `go test {...} -clean` command from within the `truss/integration_tests/` directory. So a potential solution would be to just have the makefile change directories, run the tests, then change back.

Alternatively, we could just get rid of the Makefile, though we'd need to document how to get the same functionality somewhere else.

hasAdamr:
With $ go test it is not possible from what I understand to get logging to happen while a test is running, the log will only print out after the test is complete. The test would have to be broken into a function for each directory if you want printing as it is happening.

Those errors are happening because go test finds the generated code before the clean has time to remove them.

PR #11, opened 2016-08-02
Expanding tests and documentation

hasLeland:
Expand service parser unit tests for thoroughness.

Add comments for documentation.


protoc-gen-truss-gokit/generator/clientarggen/client_argument_generator.go
```diff
@@ -1,11 +1,59 @@
...
+// does is take some arguments on the command line, convert those to the
+// correct types, place those arguments into a struct of the correct type, then
+// make a request to the server. Determining the nature of each command line
+// argument, how to convert it to the correct type, and other bookkeeping to
+// automatically generate a client is the task of the package clientarggen.
+//
+// Note that the code generation portion of creating client business logic is
+// largely done in the template files, but that those template files rely on
+// information which is collected and made conveniently available ahead of time
+// by this package.
```
hasAdamr: This seems overly verbose. You provide a large example but don't really use it and it takes a huge amount of space in the godoc. You also mention server business logic, which is unrelated to what this does.   

I feel this could be replaced with

```
// Package clientarggen marshals command line flags into golang protobuf messages fields for all rpc parameters in a gRPC service.
// Functions in clientargen are called by templates in protoc-gen-truss-gokit/template/
```

Most package descriptions I have seen in the standard library are pretty short, I feel this one could be much shorter.

hasLeland: I think you're correct that this is overly verbose, but I'm hesitant to go with the specific description of:

> Package clientarggen marshals command line flags into golang protobuf messages fields for all rpc parameters in a gRPC service.

Because it does not do that. Clientarggen is a collection of utility methods and pre-computed fields for templating code, and that generated code is what does the actual marshaling at runtime of command line arguments.

I'm going to get rid of all the verbose text and basically put that there.

hasAdamr:
```
func TestTrailingCommentsThreeDeep(t *testing.T) {
func TestTrailingCommentsTwoDeep(t *testing.T) {
func TestTrailingCommentsOneDeep(t *testing.T) {
func TestMultipleRpc(t *testing.T) {
...
```

By these test names I would not imagine that they would break if the `body` attribute got parsed wrong. So having `got, want` for every part of the parsing in every test no longer makes these unit tests, as each one seems to test everything. 

To make these unit tests, I would say there should be tests like:

```
func TestServiceName(t *testing.T) {
func TestMethodName(t *testing.T) {
func TestMethodRequestAndResponse(t *testing.T) {
func TestAdditionalBindings(t *testing.T) {
...
```

That way, if one of these features breaks then the name of the test that fails indicates what broke, rather than all the tests failing.

We probably would like to test `TestTrailingCommentsThreeDeep` and not have the test fail if the RequestType's name is wrong. Though we would want `TestMethodRequestAndResponse` to fail in that case.

I am fine with merging these tests in, though it looks like most of these tests will fail if many different things fail. Making them less useful for debugging, only really telling us if things are on fire, just not what specifically is on fire.

protoc-gen-truss-gokit/generator/clientarggen/client_argument_generator.go
hasAdamr: Maybe
Package clientarggen provides functionality for templates to marshals command line flags into golang protobuf messages fields for all rpc parameters in a gRPC service.

Though cool, looking forward to it.

hasLeland:
I'd originally written these not so much as unit tests, but as durability or functionality tests, where odd combinations of things (e.g. comments in strange places) could cause things to fail in unexpected ways. Each of these tests exists because it caught a particular bug at some point, so I don't really want to throw them away.

However, pure unit tests, like the one's you've illustrated, should be written in addition to these. I think that will be a future pull request : )

hasAdamr:
\ʕ◔ϖ◔ʔ/

PR #12, opened 2016-08-02
Update truss .go templates to .gotemplate

hasAdamr:
Having the template files be of extention .go confuses golangs tooling.
.gotemplate does not.


hasLeland:
Doesn't crash things, works well, LGTM!

hasIan:
I like appending `.tmpl` so that the actual file name is preserved. E.g., `main.go.tmpl`. But w/e.

PR #13, opened 2016-08-03
Sanitize protobuf files

hasLeland:
Remove all references to all Tune projects and products such as `mv...` or `tmc...`.


truss/integration_tests/test_service_definitions/otcentitiespb/account_user_api_key.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package tmcentitiespb;
+package otcentitiespb;
```
hasIan: What is OTC Entities?

truss/integration_tests/test_service_definitions/otcentitiespb/site.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package tmcentitiespb;
+package otcentitiespb;
 
 message SiteGetByIdRequest {
```
hasIan: I know this is protobuf, but since it ultimately generates Go, I'll note here that Go convention dictates that initialisms be all caps (like Python) rather than Java-style initial caps. So, `ID` not `Id`, `URL` not `Url`, `JSON` not `Json`, etc. (Granted "ID" is not an initialism, but it seems to follow the rule of initialisms in Go.)

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

hasIan:
I'm curious about the convention of suffixing the name of all packages in the test service definitions with `pb`. I'm imagining this carries over into the generated code, or no? If so, that's a smell. Encoding the source of the protocol/service/message definition into an output artifact is broken encapsulation.

If it doesn't carry over, then the resulting package names should be fine, but that means there's magic involved in undecorating protobuf package names, which flies in the face of Go's general avoidance of magic and adherence to the explicit.

hasLeland:
@hasIan that's a bunch of really good points. Since all of these are taken from the multiverse set of protobufs, we didn't write them so we don't know why they're in here. The `pb` suffix does carry into the generated go files, and it does feel like a code smell, so I'm going to attempt to remove them.

hasIan:
Right on; I figured it fit pretty well into a PR like this, since you're doing cleanup already. 👍  After you tend to that, you have my implied LGTM; merge away. :)

hasLeland:
Alright, the `pb` suffix has been removed from all the files.

hasIan:
So so dope. 🐋

PR #15, opened 2016-08-03
Add build flag for truss

hasAdamr:
The slowest part of truss is exec'ing $ go build for the generated
microservice. This commit adds a build flag to truss.

If truss is called as followed:
$ truss -build=false *.proto

The generated microservice will not be built.
This was added for using truss for tests that do not involve the built
microservice.


hasLeland:
Excellent!

PR #16, opened 2016-08-03
Update truss flag -build=false to -nobuild

hasAdamr:


truss/main.go
```diff
@@ -74,7 +74,7 @@ func main() {
 
 	generator.GenerateMicroservice(genImportPath, workingDirectory, definitionFiles)
 
-	if *buildMicroservice {
+	if !*noBuild {
```
hasIan, line 77: Heh, this looks really weird, but it makes the user interface more sensible. "If not no build."

truss/main.go
```diff
@@ -21,7 +21,7 @@ func init() {
 
 // Stages are documented in README.md
 func main() {
-	buildMicroservice := flag.Bool("build", true, "Set -build=false to not build microservice")
+	noBuild := flag.Bool("nobuild", false, "Set -nobuild to not build generated microservice")
```
hasIan: "to not build" sounds a little funky; I'd probably say something like "to generate code without building," or something like that.

hasIan:
Just one minor comment on the help text; address it however you see fit, and LGTM. 👍 :shipit:

truss/main.go
hasAdamr, line 77: Yeah, this is actually the reason I switched to -build=false. I did not like how the code looked.

PR #18, opened 2016-08-08
Makedt refactor

hasLeland:
Refactors the `makedt` package to be much less monolithic, moving the previous single huge `New()` method into five smaller methods which do the same things, but are much smaller and much easier to test.

Additionally, vendor the `github.com/pkg/errors` package, as it's now used within `makedt`.


hasAdamr:
Curious why some of the methods are exported and some of them are not, at a quick glance it looks like nothing but `New` "needs" to be exported. Could you say a sentence or two why `New*` is exported?

gendoc/doctree/makedt/build_doctree.go
```diff
@@ -40,7 +46,9 @@ func findDoctreePackage(req *plugin.CodeGeneratorRequest) string {
 	return ""
 }
 
-// Finds a message given a fully qualified name to that message.
+// Finds a message given a fully qualified name to that message. The provided
+// path may be either a fully qualfied name of a message, or just the bare name
+// for a message.
 func findMessage(md *doctree.MicroserviceDefinition, new_file *doctree.ProtoFile, path string) (*doctree.ProtoMessage, error) {
```
hasAdamr: Does not need to be in this pull request: A test for this function would be nice, I currently cant understand what it does at I don't know what possible inputs for "path" are. But a test would give some insight.

gendoc/doctree/makedt/build_doctree.go
```diff
@@ -1,3 +1,8 @@
+// Makedt is a package for exposing the creation of a doctree structure.
+//
+// It lives in it's own package because it must use several other packages
+// which make use of doctree to create a doctree, so to prevent circular
+// imports, it must be its own package.
```
hasAdamr, line 5: I'm a little confused about this, though you can just explain it to me in person. I'm sure it would like 15 seconds with a diagram.

gendoc/doctree/makedt/build_doctree.go
```diff
@@ -1,3 +1,8 @@
+// Makedt is a package for exposing the creation of a doctree structure.
+//
+// It lives in it's own package because it must use several other packages
```
hasIan: I've seen this a few times in the code base:

> it's = it is
> its = possessive form of it

gendoc/doctree/makedt/build_doctree.go
hasIan: Go convention prefers `camelCase` to `snake_case`. (`new_file`)

gendoc/doctree/makedt/build_doctree.go
```diff
@@ -80,92 +89,141 @@ func New(req *plugin.CodeGeneratorRequest) (doctree.Doctree, error) {
...
-				// Add values to this enum
-				n_val := doctree.EnumValue{}
-				n_val.SetName(val.GetName())
-				n_val.Number = int(val.GetNumber())
-				new_enum.Values = append(new_enum.Values, &n_val)
-			}
-			new_file.Enums = append(new_file.Enums, &new_enum)
+		newFile, err := NewFile(file, &dt)
+		if err != nil {
+			return nil, errors.Wrapf(err, "file creation of '%s' failed", file.GetName())
```
hasIan: This is a good opportunity for `%q` since we know `File.GetName` is going to return a string.

gendoc/doctree/makedt/build_doctree.go
```diff
@@ -80,92 +89,141 @@ func New(req *plugin.CodeGeneratorRequest) (doctree.Doctree, error) {
...
-				if new_field.Type.Name == "" {
-					new_field.Type.Name = field.Type.String()
-				}
-				// The label we get back is a number, translate it to a human
-				// readable string
-				label := int32(field.GetLabel())
-				label_name := descriptor.FieldDescriptorProto_Label_name[label]
-				new_field.Label = label_name
+	// Do the association of comments to units code. The implementation of this
+	// function is in `associate_comments.go`
```
hasIan: Note that GoDoc is not markdown, just plain text.

gendoc/doctree/makedt/build_doctree.go
```diff
@@ -80,92 +89,141 @@ func New(req *plugin.CodeGeneratorRequest) (doctree.Doctree, error) {
...
+	pfile *descriptor.FileDescriptorProto,
+	curNewDt *doctree.MicroserviceDefinition) (*doctree.ProtoFile, error) {
+
+	newFile := doctree.ProtoFile{}
+	newFile.Name = pfile.GetName()
+
+	for _, enum := range pfile.EnumType {
+		newEnum, err := NewEnum(enum)
+		if err != nil {
+			return nil, errors.Wrap(err, "error creating doctree.ProtoEnum")
```
hasIan: Will a truss user know what a `doctree.ProtoEnum` is? This smells of an error message written by the person writing the code. (Which is pretty common, of course, but also a trap where one writes error messages for himself and not for the audience that may encounter them in the wild.)

gendoc/doctree/makedt/build_doctree.go
```diff
@@ -80,92 +89,141 @@ func New(req *plugin.CodeGeneratorRequest) (doctree.Doctree, error) {
...
+// NewMessage returns a *doctree.ProtoMessage created from a *descriptor.DescriptorProto
+func NewMessage(msg *descriptor.DescriptorProto) (*doctree.ProtoMessage, error) {
+	newMsg := doctree.ProtoMessage{}
+	newMsg.Name = *msg.Name
+	// Add fields to this message
+	for _, field := range msg.Field {
+		newField := doctree.MessageField{}
+		newField.Number = int(field.GetNumber())
+		newField.Name = *field.Name
+		newField.Type.Name = field.GetTypeName()
```
hasIan: This looks like a candidate for an extracted function, to get me the correct type name no matter what, and to avoid a lengthy explanation in the middle of this `for` loop, where I can read `GetTypeName(...)` and know what it does and continue reading without knowing the specifics of how the type name is gotten.

gendoc/doctree/makedt/build_doctree.go
```diff
@@ -80,92 +89,141 @@ func New(req *plugin.CodeGeneratorRequest) (doctree.Doctree, error) {
...
+
+		newMsg.Fields = append(newMsg.Fields, &newField)
 	}
+	return &newMsg, nil
+}
 
-	// Do the association of comments to units code. The implementation of this
-	// function is in `associate_comments.go`
-	doctree.AssociateComments(&dt, req)
+// NewService creates a new *doctree.ProtoService from a
```
hasIan, line 186: Prefer comments that tell me what a function actually does versus sort of reiterating the parameters and return type. I can see that it takes a `*descriptor.ServiceDescriptorProto` and returns a `*doctree.ProtoService`, but this means very little to someone reading the docs for the first time.

hasLeland:
@hasAdamr I chose to export all those methods since I've been in several situations where I wished those methods existed and where public. For example, when testing other packages it was annoying to manually create a doctree or part of a doctree. In such cases I'd have greatly preferred being able to specify a ProtoDescriptor object, or some sub object, using the terse string syntax that `golang/protobuf` allows, then to use a function like these to convert that Protobuf struct to a doctree struct.

hasAdamr:
LGTM

PR #20, opened 2016-08-11
Http transport -- DO NOT MERGE

hasLeland:
DO NOT MERGE

This pull request exists purely for code review reasons. Rip this code apart!

Summary:

Client and server can now communicate with each other over HTTP. Parameters can be specified within the query and the path, and they'll be transmitted in the query and path. The body still works too.

Also, I re-wrote how the internals of several types within `doctree` work so that they're easier to work with in external packages. However, nothing else about them changed.


hasLeland:
I already remember that I need to vendor in some new dependencies, so that will happen later. The code still stands though.

hasLeland:
closing this because it's for the wrong branch.

PR #25, opened 2016-08-19
Update truss to use NamedReadWriter Interface

hasAdamr:
The truss.NamedReadWriter interface has the methods Read, Write, and
Name. *os.File fulfills this interface. A slice of them can represent a
file structure. This allows files to be directly used as NameReadWriter's.

This also allows for a truss.NamedReadWriter to be passed to functions
that accept io.Reader or io.Writer.


hasLeland:
Looks like a big improvement, and all tests pass for me. Merge if you think it's solid!

PR #32, opened 2016-08-20
Fix bug with title of services, clean repeat code

hasLeland:
Fixed a bug within the markdown generation for ProtoService where the wrong text was placed in the title of each Service. Additionally, I went through and cleaned up several cases where types had implemented their describeMarkdown methods via copy-pasted code so instead they now use a genericized genericDescribeMarkdown function which improves code clarity.


hasAdamr:
LGTM

PR #35, opened 2016-08-21
Add google-provided include files to work with binary protoc

hasLeland:
Protoc needs to know about these files, and in a binary installation of protoc, protoc doesn't necessarily know where those default protobuf files are located. So, to make sure protoc knows where these files are without us having to scrape the include path to find them and specify them on the command line, we will just include these files within truss.

So now, you can install `protoc` via a binary without worrying about the location of the `include/` folder, since those files are now included within each microservice generated via truss.


hasAdamr:
LGTM

PR #37, opened 2016-08-23
Add basic openapi/swagger generation - Maybe merge

hasAdamr:
So I am leaving today for a few weeks, and I had not gotten this finished.

This includes basic swagger generation.

Feel free to break it apart and/or clone it, fix it, re-pull request it.

It also contains a truss testing package I was working on.
The testing package `./testproto` works like this: 
1. Copy `./testproto/testprotodata` to where you want to run tests.
2. Add or change `.proto` files in `./testprotodata/definitions/` for things you want to test (.proto file with repeated message fields for example`
3. Run `$ go generate ./...`
4. Import your `testprotodata` package that was generated in the `./testprotodata` directory
5. Import `github.com/TuneLab/go-truss/testproto` 
6. Hook up `testproto` with your `testprotodata` i.e. `testData = testproto.New(testprotodata.Asset)`
7. Get `doctree` service, methods, etc with `testproto`
8. Run tests on those!


genswagger/testprotodata/generate-test-data.sh
```diff
@@ -0,0 +1,23 @@
+#! /bin/bash
+
+function gentestdata {
+	protoc -I/usr/local/include -I. -I/home/adamryman/projects/go/src/github.com/TuneLab/gob/testproto/third_party/googleapis --truss-protocast_out=./data/data $1
```
hasLeland, line 4: Hard coding a path to your home directory is maybe not the most portable thing to do ¯\_(ツ)_/¯

hasLeland:
On the topic of the testing helper library, I'd really, REALLY prefer a solution that does not require the use of a `go generate`, as I've been bit a tremendous number of times by mistakenly forgetting a `go generate` while working on a feature and having that mistake lead to bugs. If possible, I'd like to stay as far away from go generate in the tooling as possible (barring the cases where we already make heavy use of it).

Additionally, I'd prefer a solution that accepts inline protobuf definitions, allowing a developer to keep the definitions within the test(s) where they are used.

Both of the above are intended to help prevent side effects and thus, bugs and mistakes from developer error.

Here's how I think a helper library that works the way I've outlined could function, with pseudo code:

```
func DeftreeGenHelper(definition string) (deftree.Deftree, error) {
    // Since the definition comes in as a string, but we need the ast from
    // protoc, we write it to a temporary file
    file := write(tmpfilename())
    // Use protocast to get the raw ast of the file (this function call
    // hides invoking protoc with protocast as a plugin, then reading the
    // file with the raw ast)
    rawAst := procast(file.Path)
    // Marshal the raw ast into a golang-compatible ast
    ast := proto.Unmarshal(rawAst)

    // Use the ast and the definition (as bytes) to get a doctree!
    dt := deftree.New(ast, bytes(definition))
    return dt
}
```

genswagger/genswagger.go
```diff
@@ -0,0 +1,284 @@
...
+
+	err = validate.Spec(doc, strfmt.Default)
+	if err != nil {
+		return errors.Wrap(err, "swagger file cannot be validated")
+	}
+
+	return nil
+}
+
+// Spec: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#swagger-object
```
hasLeland, line 59: This comment applies to most all of the functions related to swagger generation:

Though you have technically provided a comment for most of these functions, I think adding explanations of what these function do, how each maps from `doctree` structs to swagger structs and the particularities of that mapping would make it much more clear what this package is doing and _why_.

With the only comments being links to the spec, you imply the "what", but the "how" and "why" is left unaddressed.

genswagger/genswagger.go
```diff
@@ -0,0 +1,284 @@
+package genswagger
```
hasLeland, line 1: The `genswagger` package should have a package level comment that at least contains the same information as is found in the README.md for the package.

genswagger/genswagger.go
```diff
@@ -0,0 +1,284 @@
...
+	case "TYPE_FIXED64":
+		return "string", "uint64", true
+	case "TYPE_FIXED32":
+		return "integer", "int64", true
+	case "TYPE_BOOL":
+		return "boolean", "boolean", true
+	case "TYPE_STRING":
+		return "string", "string", true
+	// case "TYPE_GROUP":
+	// case "TYPE_MESSAGE":
```
hasLeland, line 248: Is this commented out code supposed to be here?

testproto/testproto.go
```diff
@@ -0,0 +1,101 @@
+package testproto
```
hasLeland, line 1: This testing package really needs comments, as well as a README-style package level comment explaining what it is, why it exists, and giving a brief rundown on how to use it.

Issue #42, opened 2016-09-09
docker build - googlemock zip file missing


cmosetick:
When executing docker build as noted in DEVELOPING.md one currently gets `unzip: invalid zip magic`. This is because the `gmock-1.7.0.zip` that get downloaded by `autogen.sh` is not an actual zip file / URL. (So curl retrieves a 404 error HTML document instead.)

There seems to have been some rearranging of project locations.
e.g.
https://googlemock.googlecode.com/files/gmock-1.7.0.zip
vs.
https://github.com/google/googletest/tree/master/googlemock

Docker build output:

```
Removing intermediate container 40cd63641b07
Step 9 : RUN ./autogen.sh &&    ./configure --prefix=/usr &&    make &&         make install
 ---> Running in 28558c243895
Google Mock not present.  Fetching gmock-1.7.0 from the web...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1582  100  1582    0     0  11029      0 --:--:-- --:--:-- --:--:-- 12359
unzip: invalid zip magic 4F44213C
The command '/bin/sh -c ./autogen.sh &&         ./configure --prefix=/usr &&    make &&         make install' returned a non-zero code: 1
```

Note that autogen.sh has already been fixed on the master branch of the new location for googlemock:
https://raw.githubusercontent.com/google/protobuf/master/autogen.sh

However, the Dockerfile is currently pointing to branch `3.0.0-beta-4`.

hasIan:
Hi Chris, thanks for the report. I'll be pushing an update soon that takes advantage of the prebuilt binaries for the now-released protobuf 3.0.0 implementation that should alleviate this issue.

cmosetick:
Thanks @hasIan I've also submitted PR #43 for you to review that fixes this issue and cleans up the current Dockerfile.

PR #43, opened 2016-09-09
Dockerfile cleanup:

cmosetick:
- point to master branch of GoogleMock to help with issue #42.
- remove redundant RUN commands / extraneous layers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,46 +1,43 @@
 # Base Docker Image
 FROM alpine:3.4
 
-RUN apk add --no-cache git
-RUN apk add --no-cache go
-RUN apk add --no-cache openssh
-RUN apk add --update --no-cache build-base autoconf automake libtool libstdc++ curl
+RUN \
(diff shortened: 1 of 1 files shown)


hasIan:
Thanks for the PR, Chris. This looks good to alleviate the immediate issue and clean up the `Dockerfile`. I've been doing refactoring of the `Dockerfile` myself to split it into an image for building truss and a separate one for running truss, with the goal of decreasing the size of the truss-run image. The current Docker image, after it downloads all the dependencies and builds protobufs, clocks in over a gigabyte, when I checked, when all we really need at runtime are a couple binaries.

So, I'll merge this in, but know that the Docker strategy for Truss will be changing somewhat in the near future. Thanks for your interest in Truss! We welcome your input.

cmosetick:
@hasIan Thanks for the merge. Yeah, the image is 1.6 GB on my system too. I agree that separate build and run images are the best strategy for something like this moving forward.

PR #44, opened 2016-09-11
Refactor Doctree - Separate doc generation from definition structure

hasLeland:
## What does this mean for me as a developer who was using `doctree`?

This means:
- The `gendoc/doctree` package is now known as the top level `deftree` package
- You no longer have to import the package `makedt` to create a `deftree`, as the `New()` function has been moved into `deftree`
- `deftree` no longer contains logic for generating documentation. Instead that functionality now lives within the `gendoc` package.
## What is it you did to refactor `doctree`?

Here's what I've done as part of my refactor of `doctree`:
1. Consolidate `makedt` and `httpopts` into `doctree`
2. Remove coupling from `gendoc/svcparse` to `gendoc/doctree` so `gendoc/svcparse` has no dependencies on `gendoc/doctree`
3. Move `doctree` from `gendoc/doctree` into a separate top level package
   called `deftree`. 
4. Move generation of markdown documentation from
   within `deftree` into `gendoc`.
5. Remove unused `describeMarkdown` methods in `deftree`


gendoc/svcparse/scanner_test.go
```diff
@@ -15,6 +15,7 @@ func cleanStr(s string) string {
 }
 
 func TestBodyDetection(t *testing.T) {
+	t.Skip()
```
zaquestion: Can you add a comment with a quick explanation

zaquestion:
LGTM other than one small comment. I'm not familiar enough with the subsystem. Does the PR complete the desired changes for the refactor? If not I think it would be good to open an issue with a proposal for subsequent changes.

gendoc/svcparse/scanner_test.go
hasLeland: While going through those tests, I found that that test doesn't actually test anything. As in, it does some stuff and logs a bunch of data, but it doesn't actually check any condition for success or failure. I wanted to reduce the volume of needless prints, so I chose to skip that test, since it can never fail.

@zaquestion any thoughts on my reasoning? Would you have done it differently?

hasAdamr:
I think svcparse should be moved from gendoc to deftree. It is used to build deftree's and I don't think it is used anywhere else.

I did this in this pull request. https://github.com/hasLeland/go-truss/pull/1 . All tests pass. 

Otherwise LGTM.

hasLeland:
@hasAdamr I think you're correct about moving svcparse, so I've merged your pull request. Once the tests have passed for bringing in those changes, I'm merging this branch.

Issue #39, opened 2016-08-30
http_transport strconv logic needs to convert wide types to narrow types


zaquestion:
```
# github.com/zaquestion/current/service/DONOTEDIT
service/DONOTEDIT/transport_http.go:136: cannot use LatitudePostLocationFromBigBrother (type float64) as type float32 in assignment
service/DONOTEDIT/transport_http.go:146: cannot use LongitudePostLocationFromBigBrother (type float64) as type float32 in assignment
service/DONOTEDIT/transport_http.go:156: cannot use AccuracyPostLocationFromBigBrother (type float64) as type float32 in assignment
service/DONOTEDIT/transport_http.go:166: cannot use AltitudePostLocationFromBigBrother (type float64) as type float32 in assignment
service/DONOTEDIT/transport_http.go:176: cannot use BearingPostLocationFromBigBrother (type float64) as type float32 in assignment
service/DONOTEDIT/transport_http.go:186: cannot use SpeedPostLocationFromBigBrother (type float64) as type float32 in assignment
```

```
LatitudePostLocationFromBigBrotherStr := queryParams["latitude"]
LatitudePostLocationFromBigBrother, err := strconv.ParseFloat(LatitudePostLocationFromBigBrotherStr, 32)
```

The above code sets the width correctly but ParseFloat always returns the widest type so we need to convert them down when appropriate. 

In this case

```
req.Latitude = float32(LatitudePostLocationFromBigBrother)
```

Documented here:
https://golang.org/pkg/strconv/

PR #45, opened 2016-09-13
 Update http_transport decode to handle 32bit types

hasAdamr:
http_transport decode now properly converts 32bit floats and int from
64bit types return from strconv. Fixes issue #39

This PR also handles uint32 and uint64 for http_transport decode and
client argument generation

Also updating .travis.yml to run tests ./deftree directory


gengokit/httptransport/httptransport.go
```diff
@@ -181,6 +180,31 @@ func createDecodeConvertFunc(f Field) string {
 	return fmt.Sprintf(fType, f.LocalName, f.LocalName+"Str")
 }
 
+// createDecodeTypeCast creates a go string that casts a 64 bit type to a 32 bit type
+// as strconv.ParseInt and ParseFloat always return the 64 bit type
+func createDecodeTypeCast(f Field) string {
```
tunezaq: missing some types https://golang.org/ref/spec#Numeric_types

```
uint8       the set of all unsigned  8-bit integers (0 to 255)
uint16      the set of all unsigned 16-bit integers (0 to 65535)
uint32      the set of all unsigned 32-bit integers (0 to 4294967295)
uint64      the set of all unsigned 64-bit integers (0 to 18446744073709551615)

int8        the set of all signed  8-bit integers (-128 to 127)
int16       the set of all signed 16-bit integers (-32768 to 32767)
int32       the set of all signed 32-bit integers (-2147483648 to 2147483647)
int64       the set of all signed 64-bit integers (-9223372036854775808 to 9223372036854775807)
```

hasAdamr: We only support the types in this table
https://developers.google.com/protocol-buffers/docs/proto3#scalar

gengokit/httptransport/httptransport.go
```diff
@@ -181,6 +180,31 @@ func createDecodeConvertFunc(f Field) string {
...
+	fType := ""
+	switch {
```
tunezaq: Only need the int32 and float32 case in here. Everything else is default

gengokit/httptransport/httptransport.go
```diff
@@ -95,6 +95,7 @@ func NewBinding(i int, meth *deftree.ServiceMethod) *Binding {
 		}
 		nField.GoType = gt
 		nField.ConvertFunc = createDecodeConvertFunc(nField)
+		nField.TypeCast = createDecodeTypeCast(nField)
```
tunezaq: Please name TypeConversion or a bunch of gophers will attack you in your sleep.

gengokit/clientarggen/client_argument_generator.go
```diff
@@ -142,12 +142,14 @@ func New(svc *deftree.ProtoService) *ClientServiceArgs {
 func createFlagConvertFunc(a ClientArg) string {
 	fType := ""
 	switch {
+	case strings.Contains(a.GoType, "uint32"):
+		fType = "%s = flag.Uint(\"%s\", 0, %s)"
```
tunezaq: Could use a raw string here

```
`%s = flag.Uint("%s", 0, %s)`
```

gengokit/clientarggen/client_argument_generator.go
```diff
@@ -168,12 +170,14 @@ func createFlagConvertFunc(a ClientArg) string {
...
+	case strings.Contains(a.GoType, "uint64"):
+		fType = "*%s"
 	case strings.Contains(a.GoType, "int32"):
 		fType = "int32(*%s)"
 	case strings.Contains(a.GoType, "int64"):
 		fType = "*%s"
-	case strings.Contains(a.GoType, "int"):
-		fType = "*%s"
 	case strings.Contains(a.GoType, "bool"):
 		fType = "*%s"
```
tunezaq: Could default these cases.

tunezaq:
LGTM, 2 small nit pick things you could fix.

PR #48, opened 2016-09-14
 Update .travis.yml to use protoc binary

hasAdamr:
Travis should now be more reliable as it no longer is using cache
It just grabs the protoc 3.0.0 binary and adds it to the path each time


hasLeland (approved):
I'm unsure how to test this, but it seems that this makes for a welcome change, lessening the time for tests to run.

While I don't know how to verify this doesn't break automated testing, I'm willing to say:

## Looks good to me

PR #51, opened 2016-09-16
Rename 'basicService' to have name of package

hasLeland:
Within each generated service, previously it was the case that the generated go struct which contained the methods implementing the defined service was named `basicService` and the function which initialized that struct was called `NewBasicService`.

This pull request changes the function to be named `NewService` and changes the name of the struct so that it reflects the name of the package.

So if you define a package named `translator`, the name of the struct which contains the methods for that service will be `translatorService`.


tunezaq (approved):
LGTM, pending travis

PR #52, opened 2016-09-17
Fix server and client handler always being overwritten

hasAdamr:
A bug introduced when truss's output was restructured was that the
server and client handlers were not being parsed. This was a
combination of truss not reading in files properly and gengokit not
searching the read in files properly. These both have been fixed.

Also added parsing of comments of server and client handler which was
an early oversight


hasLeland:
# Looks good to me!

PR #54, opened 2016-09-17
 Update server handlers to use pointers to pb structs

hasAdamr:
grpc golang structures were being dereferenced in endpoints.go then
passed to the server handlers, then the return value was being
referenced in endpoints.go and returned up the stack. Now the pointer
is directly passed to the server handler and the server handler returns
a pointer to the response type.

Also added listing the response type message fields as comments to the server
handler template. This allows for quickly creating a response struct by
uncommenting the fields you want to populate


hasLeland:
## Looks good to me!

PR #56, opened 2016-09-20
Split Docker images into build and run images

hasIan:
Previously, there was a single Dockerfile defined that built Truss and allowed
users to run it. This yielded a tremendously large Docker image, since much was
downloaded and the entire Protobuf C++ implementation was built inside it.

To reduce the size of the Docker image that most people will end up using, one
Dockerfile became two: one that builds an image with everything needed to build
Truss, and another that just runs Truss. When the build image is run, the Truss
source should be mounted into the container at
/go/src/github.com/TuneLab/go-truss, and the script that gets run by the build
image will build Truss into a build/ directory and also copy Truss' runtime
dependencies into build/.

Once this is complete, a runtime Truss image can be generated from the contents
of the build/ directory. The runtime image weighs in at about 25 MB, a fraction
of the 1.6 GB combined image it replaces. The build image is still around a
gigabyte, but most users will not need to build Truss.

Note that we will need `tunelab/truss-build` and `tunelab/truss` repos on
Docker Hub, and it looks like I only have permission to create repos as my
personal user.


docker/build/Dockerfile
```diff
@@ -0,0 +1,61 @@
+# Truss
+# truss-build Docker image
+#
+# NOTE: YOU WILL PROBABLY NEVER NEED TO CREATE A TRUSS-BUILD IMAGE. The primary use
+# of the truss-build image will be to create updated truss-run images A pre-built
```
hasAdamr: Missing a period.

Makefile
```diff
@@ -0,0 +1,41 @@
...
+DOCKER_BUILD_IMAGE_NAME := tunelab/truss-build
+DOCKER_RUN_IMAGE_NAME := tunelab/truss
+NOW := $(shell date -u "+%Y-%m-%dT%H_%M_%S")
+
+# Default target builds Truss in using the truss-build container, then
+# packages the output into a new truss image, and removes the build output
+# from the working directory. Note that if make targets are evaluated in
+# parallel, this target will not work since it depends on the order of the
+# dependent targets as specified.
+default: docker-build build-docker-run
```
hasAdamr: I feel that by default this should build truss itself as this Makefile lives in the root of the repository.

```
default: 
    $(MAKE) -C truss
```

hasAdamr (commented):
You cannot build the truss image on linux it seems. The `tunelab/truss-build` image can be built, but when run it creates a `build` dir that is owned by root. On Mac, the `build` dir that is created is owned by the user.

Though if the `tunelab/truss` image is built on a mac, it can be pulled from dockerhub and used on linux.

Other large thing was having this Makefile at the root of the repo feels weird, it should build truss, not the docker container. Either making the default build truss or moving this Makefile to the docker folder seem like good alternatives.

hasIan:
- Running `tunelab/truss-build` image on Linux should now generate output in the `build` directory with the correct user/group, not `root:root`. Tested on Travis.
- Updated Travis to use latest versions of Go, 1.7.1 and 1.6.3 (was 1.6 and 1.7), and latest version of Protobuf, 3.0.2 (was 3.0.0).
- Addressed comment about `Makefile` organization.

hasAdamr:
We have been re-basing feature branches into one commit, @tunezaq might get upset if you don't. 

`$ docker run --rm -v $PWD:/truss tunelab/truss *.proto` runs truss, but the directories and files created are owned by root.

hasIan:
I'm not always a fan of squashing commits; it often creates artificial cohesion between changes that are not necessarily related, since in practice we don't always restrict a set of changes in a PR to be laser focused on a single change with no "hidden," unrelated changes riding along. Tightly-focused, single-change pull requests where squashing commits should always be an ideal to aspire to, even if we can't practically achieve that kind of cleanliness consistently. So I'll squash these commits even though there are a couple unrelated commits in the mix here.

However, none of this matters right now, since @hasAdamr just found that the output of the `tunelab/truss` run on Linux is afflicted with the same owned-by-root issue that the build run was. Derp; I should have thought of this. I'll add the wrapper to the `tunelab/truss` image build and run to do much the same thing as these changes to the `tunelab/truss-build` image creation and run.

hasAdamr:
Continued on pull #69

PR #57, opened 2016-09-23
Add deftree.NewFromString, Refactor gengokit. Add gengokit_test.go

hasAdamr:
`deftree` has a new function `NewFromString` which allows you to create a deftree from just a string literal, which makes testing features against in-line .proto files possible. 

gengokit.go has been refactored in order to use the function to unit test it. So much has been rewritten that the diff makes no sense. Just look at the [new file.](https://github.com/hasAdamr/go-truss/blob/testing/gengokit/gengokit.go)

Unit tests for gengokit.go have been written that test template path renaming, template rendering by path, and rerunning truss with the same definition for server_handler.go. Plus a few more.

### DIFF
--- a/deftree/build_deftree.go
+++ b/deftree/build_deftree.go
@@ -82,11 +82,14 @@ func NewFromString(def string) (Deftree, error) {
 	const defFileName = "definition.proto"
 
 	protoDir, err := ioutil.TempDir("", "truss-deftree-")
+	if err != nil {
+		return nil, errors.Wrap(err, "could not create temp directory to store proto definition")
+	}
 	defer os.RemoveAll(protoDir)
 
-	err = ioutil.WriteFile(protoDir+"/"+defFileName, []byte(def), 0666)
+	err = ioutil.WriteFile(filepath.Join(protoDir, defFileName), []byte(def), 0666)
 	if err != nil {
-		return nil, errors.Wrap(err, "could not write proto definition to filesystem")
+		return nil, errors.Wrap(err, "could not write proto definition to file")
 	}
 
 	err = protostage.Stage(protoDir)
@@ -221,7 +224,7 @@ func findMessage(md *MicroserviceDefinition, newFile *ProtoFile, path string) (*
 			}
 		}
 	}
-	return nil, fmt.Errorf("couldn't find message.")
+	return nil, fmt.Errorf("couldn't find message")
 }
 
 // NewService creates a new *ProtoService from a
--- a/deftree/build_deftree.go
+++ b/deftree/build_deftree.go
@@ -6,17 +6,19 @@ package deftree
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
 
-	"github.com/TuneLab/go-truss/deftree/svcparse"
-
 	log "github.com/Sirupsen/logrus"
 	"github.com/golang/protobuf/protoc-gen-go/descriptor"
 	plugin "github.com/golang/protobuf/protoc-gen-go/plugin"
 	"github.com/pkg/errors"
+
+	"github.com/TuneLab/go-truss/deftree/svcparse"
+	"github.com/TuneLab/go-truss/truss/protostage"
 )
 
 func init() {
@@ -29,47 +31,6 @@ func init() {
 
 	// Only log the warning severity or above.
 	log.SetLevel(log.InfoLevel)
-}
-
-// Finds the package name of the proto files named on the command line
-func findDeftreePackage(req *plugin.CodeGeneratorRequest) string {
-	for _, cmdFile := range req.GetFileToGenerate() {
-		for _, protoFile := range req.GetProtoFile() {
-			if protoFile.GetName() == cmdFile {
-				return protoFile.GetPackage()
-			}
-		}
-	}
-	return ""
-}
-
-// Finds a message given a fully qualified name to that message. The provided
-// path may be either a fully qualfied name of a message, or just the bare name
-// for a message.
-func findMessage(md *MicroserviceDefinition, newFile *ProtoFile, path string) (*ProtoMessage, error) {
-	if path[0] == '.' {
-		parts := strings.Split(path, ".")
-		for _, file := range md.Files {
-			for _, msg := range file.Messages {
-				if parts[2] == msg.GetName() {
-					return msg, nil
-				}
-			}
-		}
-		for _, msg := range newFile.Messages {
-			if parts[2] == msg.GetName() {
-				return msg, nil
-			}
-		}
-	} else {
-		for _, msg := range newFile.Messages {
-			if path == msg.GetName() {
-				return msg, nil
-			}
-		}
-	}
-	return nil, fmt.Errorf("couldn't find message.")
-
 }
 
 // New accepts a Protobuf plugin.CodeGeneratorRequest and the contents of the
@@ -115,6 +76,47 @@ func New(req *plugin.CodeGeneratorRequest, serviceFile io.Reader) (Deftree, erro
 	}
 
 	return &dt, nil
+}
+
+func NewFromString(def string) (Deftree, error) {
+	const defFileName = "definition.proto"
+
+	protoDir, err := ioutil.TempDir("", "truss-deftree-")
+	defer os.RemoveAll(protoDir)
+
+	err = ioutil.WriteFile(protoDir+"/"+defFileName, []byte(def), 0666)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not write proto definition to filesystem")
+	}
+
+	err = protostage.Stage(protoDir)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to prepare filesystem for generating deftree")
+	}
+
+	req, svcFile, err := protostage.Compose([]string{defFileName}, protoDir)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get CodeGeneratorRequest for generating deftree")
+	}
+
+	deftree, err := New(req, svcFile)
+	if err != nil {
+		return nil, errors.Wrap(err, "can not create new deftree")
+	}
+
+	return deftree, nil
+}
+
+// Finds the package name of the proto files named on the command line
+func findDeftreePackage(req *plugin.CodeGeneratorRequest) string {
+	for _, cmdFile := range req.GetFileToGenerate() {
+		for _, protoFile := range req.GetProtoFile() {
+			if protoFile.GetName() == cmdFile {
+				return protoFile.GetPackage()
+			}
+		}
+	}
+	return ""
 }
 
 // Build a new deftree.File struct
@@ -192,6 +194,34 @@ func NewMessage(msg *descriptor.DescriptorProto) (*ProtoMessage, error) {
 		newMsg.Fields = append(newMsg.Fields, &newField)
 	}
 	return &newMsg, nil
+}
+
+// Finds a message given a fully qualified name to that message. The provided
+// path may be either a fully qualfied name of a message, or just the bare name
+// for a message.
+func findMessage(md *MicroserviceDefinition, newFile *ProtoFile, path string) (*ProtoMessage, error) {
+	if path[0] == '.' {
+		parts := strings.Split(path, ".")
+		for _, file := range md.Files {
+			for _, msg := range file.Messages {
+				if parts[2] == msg.GetName() {
+					return msg, nil
+				}
+			}
+		}
+		for _, msg := range newFile.Messages {
+			if parts[2] == msg.GetName() {
+				return msg, nil
+			}
+		}
+	} else {
+		for _, msg := range newFile.Messages {
+			if path == msg.GetName() {
+				return msg, nil
+			}
+		}
+	}
+	return nil, fmt.Errorf("couldn't find message.")
 }
 
 // NewService creates a new *ProtoService from a
--- a/deftree/build_deftree_test.go
+++ b/deftree/build_deftree_test.go
@@ -14,6 +14,47 @@ import (
 	descriptor "github.com/golang/protobuf/protoc-gen-go/descriptor"
 )
 
+func TestNewFromString(t *testing.T) {
+	const def = `
+		syntax = "proto3";
+
+		// General package
+		package general;
+
+		import "google/api/annotations.proto";
+
+		// RequestMessage is so foo
+		message RequestMessage {
+		  string input = 1;
+		}
+
+		// ResponseMessage is so bar
+		message ResponseMessage {
+		  string output = 1;
+		}
+
+		// ProtoService is a service
+		service ProtoService {
+		  // ProtoMethod is simple. Like a gopher.
+		  rpc ProtoMethod (RequestMessage) returns (ResponseMessage) {
+			// No {} in path and no body, everything is in the query
+			option (google.api.http) = {
+			  get: "/route"
+			};
+		  }
+		}
+	`
+
+	deftree, err := NewFromString(def)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if got, want := deftree.GetName(), "general"; got != want {
+		t.Errorf("\n`%v` was deftree package name\n`%v` was wanted", got, want)
+	}
+}
+
 func TestNewFile(t *testing.T) {
 	src := `
 		name: "path/to/example.proto",
--- a/deftree/build_deftree_test.go
+++ b/deftree/build_deftree_test.go
@@ -51,7 +51,7 @@ func TestNewFromString(t *testing.T) {
 	}
 
 	if got, want := deftree.GetName(), "general"; got != want {
-		t.Errorf("\n`%v` was deftree package name\n`%v` was wanted", got, want)
+		t.Errorf("`%v` was deftree package name \n`%v` was wanted", got, want)
 	}
 }
 
--- a/gengokit/gengokit.go
+++ b/gengokit/gengokit.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"go/format"
 	"io"
-	//"io/ioutil"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
@@ -31,10 +31,6 @@ func init() {
 	})
 }
 
-type generator struct {
-	templateExec templateExecutor
-}
-
 // templateExecutor is passed to templates as the executing struct its fields
 // and methods are used to modify the template
 type templateExecutor struct {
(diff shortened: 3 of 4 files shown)


gengokit/gengokit.go
```diff
@@ -86,37 +117,103 @@ func getProtoService(dt deftree.Deftree) (*deftree.ProtoService, error) {
...
-			ImportPath:  importPath,
-			PackageName: pkgName,
-			Service:     service,
-			ClientArgs:  clientarggen.New(service),
-			HTTPHelper:  httptransport.NewHelper(service),
-		},
+// generateResponseFile contains logic to choose how to render a template file
+// based on path and if that file was generated previously. It accepts a
+// template path to render, a templateExecutor to apply to the template, and a
+// map of paths to files for the pervious generation. It returns a
```
hasLeland: "previous" is miss-spelled as "pervious".

gengokit/gengokit.go
```diff
@@ -160,134 +265,64 @@ func (g *generator) updateClientMethods(clientHandler io.Reader, svcFuncs []stri
...
-			continue
-		}
-
-		var generatedFilePath string
-		var generatedCode *bytes.Buffer
-		var err error
-
-		if templateFilePath == "NAME"+serverHandlerFilePath+"template" && serverHandlerFile != nil {
-			// If there's an existing service file, update its contents
+	svcFuncs = append(svcFuncs, "NewService")
```
hasLeland: While I may understand exactly why this code is manually adding "NewService" here, there should probably be a comment in the code explaining this behavior.

gengokit/gengokit.go
```diff
@@ -298,20 +333,14 @@ func applyTemplateFromPath(templFilePath string, executor interface{}) (*bytes.B
...
-	funcMap := template.FuncMap{
-		"ToLower":    strings.ToLower,
-		"Title":      strings.Title,
-		"GoName":     generatego.CamelCase,
-		"TrimPrefix": strings.TrimPrefix,
-	}
-
-	codeTemplate, err := template.New(templName).Funcs(funcMap).Parse(templateString)
+	// funcMap is defined at the top of this file
+	codeTemplate, err := template.New(templName).Funcs(executor.funcMap).Parse(templateString)
```
hasLeland: The comment above says "funcMap is defined at the top of this file", which seems to imply that `funcMap` is an unexported global, but `funcMap` is simply a field on the executor, and since it's a field it's evident "by inspection" (as in, just by looking at it) that if I want to know what `funcMap` is all about, I can find it by looking at the definition of `executor`. So the comment above should probably be removed.

gengokit/gengokit.go
```diff
@@ -47,28 +43,63 @@ type templateExecutor struct {
...
+		funcMap:     funcMap,
+	}, nil
+}
+
+// GenerateGokit accepts a deftree representing the ast of a group of .proto
+// files, a []truss.NamedReadWriter representing files generated previously,
+// and a goImportPath for templating go code imports. GenerateGokit returns the
+// a []truss.NamedReadWriter representing a generated gokit service file
+// structure
+func GenerateGokit(dt deftree.Deftree, previousFiles []truss.NamedReadWriter, goImportPath string) ([]truss.NamedReadWriter, error) {
```
hasLeland: The docstring should mention that `previousFiles` can be nil if there are no previous files.

gengokit/gengokit.go
```diff
@@ -160,134 +265,64 @@ func (g *generator) updateClientMethods(clientHandler io.Reader, svcFuncs []stri
...
-		var resp truss.SimpleFile
+	return svcFuncs
+}
 
-		// Switch "NAME" in path with packageName.
-		//i.e. for packageName = addsvc; /NAME-service/NAME-server -> /addsvc-service/addsvc-server
-		resp.Path = strings.Replace(generatedFilePath, "NAME", packageName, -1)
+// trimTemplateExecutorServiceFuncs removes functions in funcsInFile from the
+// passed templateExecutor and returns a pointer to that templateExecutor
+func trimTemplateExecutorServiceFuncs(te templateExecutor, funcsInFile map[string]bool) (*templateExecutor, error) {
```
hasLeland: This function only makes use of the `Service` field of `te`, so it shouldn't take an entire templateExecutor, only a pointer to a service.

gengokit/gengokit.go
```diff
@@ -160,134 +265,64 @@ func (g *generator) updateClientMethods(clientHandler io.Reader, svcFuncs []stri
...
+	// Service's pointer to it's messages to be methodsToTemplate
 
-		codeGenFiles = append(codeGenFiles, &resp)
-	}
+	tempService := *te.Service
+	tempService.Methods = methodsToTemplate
 
-	return codeGenFiles, nil
-}
+	te.Service = &tempService
```
hasLeland: `te.Service.Methods` is a pointer to `methodsToTemplate`, so ranging over `t.Service.Methods` and modifying `methodsToTemplate` means you're ranging over the thing you're modifying, which is bad and will probably break things. Fix ur pointers, yo!

hasLeland (changes requested):
Some comment edits, but also a bug/pointer confusion. Additionally, `templateExecutor` currently feels a bit like a god-object that's passed around over-zealously, with state being mutated in odd ways (instead of state being discretely re-built, then over-written). I will return (if that would be helpful) with more specific suggestions and examples of how things are confusing right now.

gengokit/gengokit.go
hasAdamr: It is commented in the method header, this method is pretty short, should it really be repeated here?

gengokit/gengokit.go
hasAdamr: But we want a new templateExector to be passed to a applyTemplate, so we want to create a copy of the templateExector passed in and return a pointer to that copy modified.

gengokit/gengokit.go
hasAdamr: Woops, that should have been below that range. Moved and now it should be fine I think. Writing a test for it now though.

hasAdamr:
`templateExecutor`is never overwritten, it is copied and a modified version is returned with the `trimTemplateExecutorServiceFuncs` function.

Before this refactor everything was a `*generator` method and so `templateExecutor` was being passed around like a god object through `generator`. Should all of these functions be `*templateExecutor` methods to avoid passing around a `*templateExecutor`?

deftree/build_deftree.go
```diff
@@ -117,6 +78,47 @@ func New(req *plugin.CodeGeneratorRequest, serviceFile io.Reader) (Deftree, erro
 	return &dt, nil
 }
 
+func NewFromString(def string) (Deftree, error) {
+	const defFileName = "definition.proto"
+
+	protoDir, err := ioutil.TempDir("", "truss-deftree-")
```
hasIan, line 84: You're assigning `err` but never check it.

deftree/build_deftree.go
```diff
@@ -117,6 +78,47 @@ func New(req *plugin.CodeGeneratorRequest, serviceFile io.Reader) (Deftree, erro
...
+	defer os.RemoveAll(protoDir)
+
+	err = ioutil.WriteFile(protoDir+"/"+defFileName, []byte(def), 0666)
```
hasIan: Don't hardcode "/". use `os.PathSeparator`.

deftree/build_deftree.go
```diff
@@ -117,6 +78,47 @@ func New(req *plugin.CodeGeneratorRequest, serviceFile io.Reader) (Deftree, erro
...
+	if err != nil {
+		return nil, errors.Wrap(err, "could not write proto definition to filesystem")
```
hasIan: "to filesystem" is a long way to say "file" here.

deftree/build_deftree.go
```diff
@@ -117,6 +78,47 @@ func New(req *plugin.CodeGeneratorRequest, serviceFile io.Reader) (Deftree, erro
...
+	}
+
+	err = protostage.Stage(protoDir)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to prepare filesystem for generating deftree")
```
hasIan, line 97: Make these error messages understandable to someone who didn't write Truss.

deftree/build_deftree.go
```diff
@@ -194,6 +196,34 @@ func NewMessage(msg *descriptor.DescriptorProto) (*ProtoMessage, error) {
...
+			}
+		}
+	} else {
+		for _, msg := range newFile.Messages {
+			if path == msg.GetName() {
+				return msg, nil
+			}
+		}
+	}
+	return nil, fmt.Errorf("couldn't find message.")
```
hasIan: Don't put a period in your error message.

deftree/build_deftree_test.go
```diff
@@ -14,6 +14,47 @@ import (
...
+		  }
+		}
+	`
+
+	deftree, err := NewFromString(def)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if got, want := deftree.GetName(), "general"; got != want {
```
hasIan, line 53: This is weird. Why not

```
if actual := deftree.GetName(); actual != "general" {
    ...
}
```

I don't care about the name; I just use expected/actual in my tests.

deftree/build_deftree_test.go
```diff
@@ -14,6 +14,47 @@ import (
...
+		t.Errorf("\n`%v` was deftree package name\n`%v` was wanted", got, want)
```
hasIan: Why embedding a leading newline in the failure message?

gengokit/gengokit.go
```diff
@@ -47,28 +43,64 @@ type templateExecutor struct {
...
-// GenerateGoCode returns the a []truss.NamedReadWriter representing a generated
-// gokit microservice file structure
-func GenerateGokit(dt deftree.Deftree, previousFiles []truss.NamedReadWriter, goImportPath string) ([]truss.NamedReadWriter, error) {
+// newTemplateExecutor accepts a deftree and a goImportPath to construct a
+// templateExecutor for templating a service
+func newTemplateExecutor(dt deftree.Deftree, goImportPath string) (*templateExecutor, error) {
 	service, err := getProtoService(dt)
 	if err != nil {
-		return nil, errors.Wrap(err, "no service found aborting generating gokit microservice")
+		return nil, errors.Wrap(err, "no service found aborting generating gokit service")
```
hasIan: Add a semicolon or something to separate "found" and "aborting." It's difficult to make sense of otherwise.

gengokit/gengokit.go
```diff
@@ -47,28 +43,64 @@ type templateExecutor struct {
...
+}
 
+// GenerateGokit accepts a deftree representing the ast of a group of .proto
+// files, a []truss.NamedReadWriter representing files generated previously
+// (which may be nil for no previously generated files), and a goImportPath for
+// templating go code imports. GenerateGokit returns the a
+// []truss.NamedReadWriter representing a generated gokit service file
+// structure
+func GenerateGokit(dt deftree.Deftree, previousFiles []truss.NamedReadWriter, goImportPath string) ([]truss.NamedReadWriter, error) {
+
```
hasIan: Don't start functions with blank lines.

gengokit/gengokit.go
```diff
@@ -86,37 +118,103 @@ func getProtoService(dt deftree.Deftree) (*deftree.ProtoService, error) {
...
+	// if no code has been generated just apply the template
+	if genCode == nil {
+		genCode, err = applyTemplateFromPath(templFP, te)
+	}
+
+	if err != nil {
+		return nil, errors.Wrap(err, "could not render template")
+	}
+
+	// Turn code buffer into string and format it
```
hasIan: This comment does not describe what happens directly beneath it. I don't think it's even necessary.

gengokit/gengokit.go
```diff
@@ -86,37 +118,103 @@ func getProtoService(dt deftree.Deftree) (*deftree.ProtoService, error) {
...
+
+	// ignore error as we want to write the code either way to inspect after
+	// writing to disk
+	formattedCode, _ := formatCode(codeBytes)
+
+	var resp truss.SimpleFile
+
+	// Set the path to the file and write the code to the file
+	resp.Path = actualFP
+	resp.Write(formattedCode)
```
hasIan: This is missing an error check.

gengokit/gengokit.go
```diff
@@ -86,37 +118,103 @@ func getProtoService(dt deftree.Deftree) (*deftree.ProtoService, error) {
...
+	return actual
 }
 
-// updateServerMethods will update the functions within an existing server_handler.go
-// file so it contains all the functions within svcFuncs
-func (g *generator) updateServerMethods(svcHandler io.Reader, svcFuncs []string) (outCode *bytes.Buffer, err error) {
+// filesTofilepathMap accepts a slice of truss.NamedReadWriters and returns a map of filepaths
+// as strings to files as io.Reader
+func filesToFilepathMap(previousFiles []truss.NamedReadWriter) (pathToFile map[string]io.Reader) {
+	pathToFile = make(map[string]io.Reader)
```
hasIan: When you know how many maps an element will hold, you can pre-initialize it. Probably not a big deal here since it will probably all fit in the initial initialization (and performance is not going to be an issue with Truss, generally speaking), but this could be

```
pathToFile = make(map[string]io.Reader, len(previousFiles))
```

However, I would get rid of the named return parameter here. You're returning it by name, so you're not using it except to eliminate a colon in the initial assignment. Generally speaking, avoid named return parameters unless you have a compelling reason to use them. They're generally less readable than variables declared inline.

gengokit/gengokit.go
```diff
@@ -125,31 +223,39 @@ func (g *generator) updateServerMethods(svcHandler io.Reader, svcFuncs []string)
...
 
-	err = g.applyTemplateForMissingMeths(svcMethodsTemplPath, currentFuncs, code)
+	trimmedTe, err := trimTemplateExecutorServiceFuncs(*te, currentFuncs)
+
+	out, err := applyTemplateFromPath(svcMethodsTemplPath, trimmedTe)
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to apply service methods template")
 	}
 
+	code.ReadFrom(out)
```
hasIan: This line makes `out` seem like a really unintuitive variable name.

gengokit/gengokit.go
```diff
@@ -125,31 +223,39 @@ func (g *generator) updateServerMethods(svcHandler io.Reader, svcFuncs []string)
...
-// updateClientMethods will update the functions within an existing
-// client_handler.go file so that it contains exactly the fucntions passed in
-// svcFuncs, no more, no less.
-func (g *generator) updateClientMethods(clientHandler io.Reader, svcFuncs []string) (outCode *bytes.Buffer, err error) {
+// updateClientMethods accepts NAME-service/handlers/client/client_handler.go
+// as an io.Reader and modifies it to contains only functions returned by
+// serviceFunctionNames.
+func updateClientMethods(clientHandler io.Reader, te *templateExecutor) (outCode io.Reader, err error) {
+	svcFuncs := serviceFunctionsNames(te.Service.Methods)
+
```
hasIan, line 257: Something about declaring constants inside a function seems weird. I imagine there's no penalty since the compiler and move them out of the actual function execution, but I typically expect constants declared at a package level, like,

```
const (
   myFirstConst = "My first constant!"
   mySecondConst = "Hey gang, this is my second const."
   myFourthConst = "What happened to the third constant?"
   myFifthConst = "We don't talk about the third constant."
)
```

I'm sort of torn, because having them declared right in the function is very readable since it maintains locality, but traditionally, if something is never going to change, it's not declared inside a function body.

gengokit/gengokit.go
```diff
@@ -160,134 +266,62 @@ func (g *generator) updateClientMethods(clientHandler io.Reader, svcFuncs []stri
...
-	// which should exist within the handler files
-	var serviceFunctions []string
-	for _, meth := range g.templateExec.Service.Methods {
-		serviceFunctions = append(serviceFunctions, meth.GetName())
+// serviceFunctionNames returns a slice of function names which are in the
+// definition files plus the function "NewService". Used for inserting and
+// removing functions from previously generated handler files
+func serviceFunctionsNames(meths []*deftree.ServiceMethod) []string {
+	var svcFuncs []string
+	for _, meth := range meths {
```
hasIan: Something about having variables named `meths` and `meth` instead of `methods` and `m` seems kind of wrong to me. In an amusing way.

gengokit/gengokit.go
```diff
@@ -160,134 +266,62 @@ func (g *generator) updateClientMethods(clientHandler io.Reader, svcFuncs []stri
...
+// passed templateExecutor and returns a pointer to that templateExecutor
+func trimTemplateExecutorServiceFuncs(te templateExecutor, funcsInFile map[string]bool) (*templateExecutor, error) {
 	var methodsToTemplate []*deftree.ServiceMethod
-	for _, meth := range g.templateExec.Service.Methods {
+
+	for _, meth := range te.Service.Methods {
 		methName := meth.GetName()
-		if funcIndex[methName] == false {
+
+		if funcsInFile[methName] == false {
```
hasIan: I'm suspicious about `funcsInFile`. If the key does not exist, then `false` will be returned since that's the zero value of a `bool`. This seems more reliable:

```
funcsInFile map[string]struct{}
...
if _, inFile := funcsInFile[methName]; inFile {
    continue
}
methodsToTemplate = append(methodsToTemplate, meth)
log.WithField("Method", methName).Info("Rendering template for method")
```

gengokit/gengokit.go
```diff
@@ -160,134 +266,62 @@ func (g *generator) updateClientMethods(clientHandler io.Reader, svcFuncs []stri
...
 		}
 	}
 
-	// Create temporary templateExec with only the methods we want to append
-	// We must also dereference the templateExec's Service and change our newly created
-	// Service's pointer to it's messages to be methodsToTemplate
-	templateExecWithOnlyMissingMethods := g.templateExec
-	tempService := *g.templateExec.Service
+	// templateExec's Service is dereference and that new Service's
+	// pointer to it's messages is changed to be methodsToTemplate
```
hasIan: "it's" means "it is."

gengokit/gengokit.go
```diff
@@ -329,14 +356,14 @@ func applyTemplate(templBytes []byte, templName string, executor interface{}) (*
...
-func formatCode(code []byte) []byte {
+func formatCode(code []byte) ([]byte, error) {
 	formatted, err := format.Source(code)
 
 	if err != nil {
 		log.WithError(err).Warn("Code formatting error, generated service will not build, outputting unformatted code")
-		// Set formatted to code so at least we get something to examine
-		formatted = code
+		// return code so at least we get something to examine
+		return code, err
```
hasIan: This is kind of unidiomatic. Typcially if a function returns an error, it returns the zero value of any other return values. The caller typically checks for an error and acts on that. The caller in this case could know that this function returns other values even when an error is returns, and inspect that, but that's a one-off that requires knowledge about implementation details of the function that is being called, which is pretty un-Go-like in that it's a sort of "magic."

gengokit/gengokit_test.go
```diff
@@ -0,0 +1,432 @@
...
+			// No {} in path and no body, everything is in the query
+			option (google.api.http) = {
+			  get: "/route"
+			};
+		  }
+		}
+	`
+	dt, err := deftree.NewFromString(def)
+	if err != nil {
+		t.Error(err)
```
hasIan: `testing.T.Error` will fail the test but not stop test execution. If you want to stop test execution, use `testing.T.Fail`.

gengokit/gengokit_test.go
```diff
@@ -0,0 +1,432 @@
...
+
+	// format that new code
+	sameCode, err = formatCode(sameCode)
+	if err != nil {
+		t.Error(err)
+	}
+
+	// make sure the code is the same
+	if bytes.Compare(shCode, sameCode) != 0 {
+		t.Fatalf("\n__BEFORE__\n\n%v\n\n__AFTER\n\n%v\n\nCode before and after updating differs",
```
hasIan: I've noticed there are a lot of `%v`s in the Truss code where there would more appropriately be `%s`s. It's not a problem in terms of execution, but it's stylistically odd.

deftree/build_deftree_test.go
hasAdamr: This was for printing "general" in the log message. I wanted to define in once place and use it in the log rather than having it hard coded in both places.

deftree/build_deftree_test.go
hasAdamr: This was for lining up the got and wanted in the printed test output.

No new lines:

```
    build_deftree_test.go:54: `general` was deftree package name `genera` was wanted
```

Only mid new line

```
    build_deftree_test.go:54: `general` was deftree package name 
            `genera` was wanted
```

Leading new line:

```
    build_deftree_test.go:54: 
        `general` was deftree package name
        `genera` was wanted
```

gengokit/gengokit.go
hasLeland: You're right, the explanation in the doc is sufficient, adding an additional comment like I suggested is unnecessary.

gengokit/gengokit.go
hasLeland: If the point of the function is to trim the service functions, then I think it should _just_ it should trim the service functions. As in, it should accept a slice of `ServiceMethod`s, then return a different slice which is then added back on to the template executor.

If the goal though is to invisibly mutate the state of the `TemplateExecutor`, then I think `trimTemplateExecutorServiceFuncs` should probably be a method on `TemplateExecutor`.

gengokit/gengokit.go
hasAdamr: Fixed, now no longer returns an error.

gengokit/gengokit_test.go
hasAdamr: This has been fixed.

gengokit/gengokit_test.go
hasAdamr: I'll try to improve this practice, fixed here and a few other places

gengokit/gengokit.go
hasAdamr: Fixed.

gengokit/gengokit.go
hasAdamr: I like that a lot more, changed.

gengokit/gengokit.go
hasAdamr: Changing this because this seems way more go.

gengokit/gengokit.go
hasAdamr: Keeping them in the function, because they feel good in the function and that's where you would go to update them.

gengokit/gengokit.go
hasAdamr: Yep, the names have been updated.

gengokit/gengokit.go
hasAdamr: Did both of these.

gengokit/gengokit.go
hasAdamr: Found a few more of those, oops.

gengokit/gengokit.go
hasAdamr: Removed

gengokit/gengokit.go
hasAdamr: Got it.

gengokit/gengokit.go
hasAdamr: Done.

deftree/build_deftree.go
hasAdamr: Got it.

deftree/build_deftree.go
hasAdamr: This will be dealt with in another pull request after this one is finished.

deftree/build_deftree.go
hasAdamr: Agreed.

deftree/build_deftree.go
hasAdamr: Now I know about filepath.Join and filepath.Rel!

deftree/build_deftree.go
hasAdamr: Got to be careful about that.

gengokit/gengokit.go
hasAdamr: Done.

gengokit/gengokit.go
hasAdamr: Fixed.

gengokit/gengokit.go
hasAdamr: It is now templateExector.trimServiceFuncs

gengokit/gengokit.go
hasAdamr: Fixed

hasLeland (approved):

deftree/build_deftree_test.go
hasIan: Ah okay, I must have misssd that. Fair use in that case.

deftree/build_deftree_test.go
hasIan: Fair enough, just kind of nonstandard and never seen that before. But I like the formatting with the newlines.

PR #60, opened 2016-09-28
Fix http transport panic

hasAdamr:
Pull request #54 introduced a bug where http transport panics. I know this from using truss. I don't know if this covers all the cases, but it does make truss usable.

I am currently adding http transport to the integration tests to add another layer of production. That pull request should be up soon.


hasIan:
I haven't run into this probably because I'm using a pre #54 Truss in my trussification branches. This change is clear though, so, LGTM. :shipit:

tunezaq (approved):

PR #62, opened 2016-09-28
Add support for repeated fields

hasLeland:
Adds support for repeated fields in http transport generation and within client argument handling.
## Client and Server Repeated Fields

Let's say you where to use our `addsvc` proto file, but set the `A` field of SumRequest to be `repeated`. This implies a slice of int64 instead of a single int64. So, on the command line, you'd type:

```
$ addsvc-client -http.addr=:8081 -sum.a "[1,3]" -sum.b 2
```

And the server would respond like so:

```
Encoding request a:1 a:3 b:2 
URL: http://:8081/sum?a=1&a=3&b=2
Client Requested with:
[1 3] 2
Server Responded with:
v:6 
```
### Repeated fields in the Path are forbidden

This does NOT add support for repeated fields within the path of an HTTP request, because:
1. There was no reliable way to do this
2. It is something that is totally unsupported by other HTTP frameworks/libraries because of reason 1

However, if you try to put a repeated field into the path, Truss will now provide a nice warning telling you not to put a repeated field into the path.
### This does NOT add support for nested messages

Additionally, this does NOT add support for arbitrarily nested messages. If you try to nest a message or use a map, Truss will now warn you about how those things are not supported.


gengokit/clientarggen/client_argument_generator.go
```diff
@@ -23,15 +23,55 @@ import (
...
-	ProtbufType     string
-	GoType          string
+	// Name contains the name of the arg as it appeared in the original
+	// protobuf definition.
+	Name string
+
+	// FlagName is the name of the command line flag to be passed to set this
+	// argument.
+	FlagName string
+	// FlagArg is the name of the Go variable that will hold the result of
```
hasIan, line 33: Use consistent vertical spacing inside a struct definition. Suggest leaving a blank line before the documentation comments, as you have after `Name`.

gengokit/clientarggen/client_argument_generator.go
```diff
@@ -109,75 +178,113 @@ func New(svc *deftree.ProtoService) *ClientServiceArgs {
...
+		return GenerateCarveInvocation(&a)
+	}
+	return fmt.Sprintf(`%s := %s`, a.GoArg, flagTypeConversion(a))
+}
+
 // createFlagConvertFunc creates the go string for the flag invocation to parse
 // a command line argument into it's nearest available type that the flag
 // package provides.
 func createFlagConvertFunc(a ClientArg) string {
 	fType := ""
```
hasIan, line 250: When you don't need to set an initial value, prefer the `var fType string` form.

gengokit/clientarggen/client_argument_generator_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/davecgh/go-spew/spew"
+
+	"github.com/TuneLab/go-truss/deftree"
+	"github.com/TuneLab/go-truss/gengokit/gentesthelper"
+)
+
+var (
+	spw = spew.ConfigState{
+		Indent: "   ",
```
hasIan, line 15: THREE spaces!? You savage!

gengokit/clientarggen/client_argument_generator_test.go
```diff
@@ -0,0 +1,112 @@
...
+					&ClientArg{
+						Name:            "a",
+						FlagName:        "sum.a",
+						FlagArg:         "flagASum",
+						FlagType:        "string",
+						FlagConvertFunc: "flagASum = flag.String(\"sum.a\", \"\", \"\")",
+						GoArg:           "ASum",
+						GoType:          "[]int64",
+						GoConvertInvoc:  "ASum := CarveASum(*flagASum)",
+						GoConvertFunc:   "\nfunc CarveASum(inpt string) []int64 {\n\tinpt = strings.Trim(inpt, \"[] \")\n\tslc := strings.Split(inpt, \",\")\n\tvar rv []int64\n\n\tfor _, item := range slc {\n\t\titem = strings.Trim(item, \" \")\n\t\titem = strings.Replace(item, \"'\", \"\\\"\", -1)\n\t\tif len(item) == 0 {\n\t\t\tcontinue\n\t\t}\n\t\ttmp, err := strconv.ParseInt(item, 10, 64)\n\t\tif err != nil {\n\t\t\tpanic(fmt.Sprintf(\"couldn't parse '%v' of '%v'\", item, inpt))\n\t\t}\n\t\trv = append(rv, tmp)\n\t}\n\treturn rv\n}\n\n",
```
hasIan, line 85: This is pretty gnarly. Generally, I try to keep lines no more than 100 characters long for readability's sake, but even neglecting this, the escape sequences impair readability. Define this as a raw string variable somewhere else and reference it in the assignment.

E.g.,

``` go
testGoConvertFunc := `
func CarveASum(input string) []int64 {
    input = strings.Trim(input, "[] ")
    slice := strings.Split(input, ",")
    var rv []int64

    for _, item := range slice {
        item = strings.Replace(item, "'", "\\", -1)
        if len(item) == 0 {
            continue
        }
        tmp, err := strconv.ParseInt(item, 10, 64)
        if err != nil {
            panic(fmt.Sprintf("couldn't parse %q of %q", item, input))
        }
        rv = append(rv, tmp)
    }
    return rv
}`

// ...
GoConvertFunc: testGoConvertFunc
```

gengokit/clientarggen/client_argument_generator_test.go
```diff
@@ -0,0 +1,112 @@
...
+						IsBaseType:      true,
+						Repeated:        true,
+					},
+					&ClientArg{
+
+						Name:            "b",
+						FlagName:        "sum.b",
+						FlagArg:         "flagBSum",
+						FlagType:        "int64",
+						FlagConvertFunc: "flagBSum = flag.Int64(\"sum.b\", 0, \"\")",
```
hasIan, line 96: Again, use a raw string here (inline is fine) to eliminate the need to use all those backslashes.

gengokit/clientarggen/client_argument_generator_test.go
```diff
@@ -0,0 +1,112 @@
...
+						GoConvertFunc:   "",
+						ProtbufType:     "TYPE_INT64",
+						IsBaseType:      true,
+						Repeated:        false,
+					},
+				},
+			},
+		},
+	}
+	if got, want := csa, expected; !reflect.DeepEqual(got, want) {
```
hasIan, line 109: I saw this pattern in one of Adam's pull request and wondered where it came from, but it's even more bizarre here. You have a variable called `expected`, but instead of using it, you're assigning it to another variable called `want`. Further, the actual result is already in a variable called `csa`. Why not

```
if !reflect.DeepEqual(expected, csa) {
// ...
```

gengokit/clientarggen/strcarver.go
```diff
@@ -0,0 +1,138 @@
...
+	code, err := ApplyTemplate("CarveTemplate", CarveTemplate, executor, nil)
+	if err != nil {
+		panic(fmt.Sprintf("Couldn't apply template: %v", err))
+	}
+
+	return code
+}
+
+func GenerateCarveInvocation(m *ClientArg) string {
+	code, err := ApplyTemplate("CarveInvocationTempl", invocTemplate, m, nil)
```
hasIan, line 56: This is an odd abbreviation. "Tmpl" or "template" seem fine, but "Templ" is like you changed your mind about abbreviating halfway through typing "template."

I realize this is pretty nitpicky.

gengokit/clientarggen/strcarver.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+		rv = append(rv, {{.TypeConv}})
+	}
+	return rv
+}
+
+`
+
+func ParseFunction(gotype string) string {
+	invoc := ""
```
hasIan, line 89: Use `var invoc string` to declare variable and assign implicit zero value.

gengokit/clientarggen/strcarver.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+		{{.ParseFunc}}
+		if err != nil {
+			panic(fmt.Sprintf("couldn't parse '%v' of '%v'", item, inpt))
+		}
+		rv = append(rv, {{.TypeConv}})
+	}
+	return rv
+}
+
```
hasIan, line 85: Generally speaking, aggressively expunge unnecessary vertical whitespace.

gengokit/clientarggen/templated.go
```diff
@@ -0,0 +1,169 @@
+package clientarggen
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+//
```
hasIan, line 9: Blank comment and extra blank line.

gengokit/clientarggen/templated.go
```diff
@@ -0,0 +1,169 @@
...
+
+	for _, item := range slc {
+		item = strings.Trim(item, " ")
+		item = strings.Replace(item, "'", "\"", -1)
+		if len(item) == 0 {
+			continue
+		}
+		tmp, err := strconv.ParseInt(item, 10, 64)
+		if err != nil {
+			panic(fmt.Sprintf("couldn't parse '%v' of '%v'", item, inpt))
```
hasIan, line 44: These are both strings; why not `"%q of %q"`?

gengokit/clientarggen/templated.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+		rv = append(rv, tmp)
+	}
+	return rv
+}
+
+func CarveUintthirtytwo(inpt string) []uint32 {
```
hasIan, line 51: What's with the numbers written out in these function names? `sixtyfour`, `thirtytwo`. What am I missing?

gengokit/clientarggen/templated.go
```diff
@@ -0,0 +1,169 @@
...
+		tmp, err := strconv.ParseFloat(item, 64)
+		if err != nil {
+			panic(fmt.Sprintf("couldn't parse '%v' of '%v'", item, inpt))
+		}
+		rv = append(rv, tmp)
+	}
+	return rv
+}
+
+func CarveBool(inpt string) []bool {
```
hasIan, line 131: These `Carve` functions really make we want a generic of some sort. If it were not for the return values of each function. I wanted to otherwise suggest passing a parsing function to each. You could still do it like so, but it's probably not worth the additional complexity.

```
func Carve(input string, parser func(string) (interface{}, error)) []interface{} {
    input = strings.Trim(input, "[]")
    slice := strings.Split(input, ",")
    var rv []interface{}
    for _, item := range slice {
        item = strings.Trim(item, " ")
        item = strings.Replace(item, "'", `\`, -1)
        if len(item) == 0 {
            continue
        }
        tmp, err := parser(item)
        if err != nil {
            panic(fmt.Sprintf("could not parse %q of %q", item, input))
        }
        rv = append(rv, tmp)
    }
    return rv
}

func CarveFloat64(input string) []float64 {
    parser := func(input string) (interface{}, error) {
        return strconv.ParseFloat(input, 64)
    }
    intf := Carve(input, parser)
    floats := make([]float64, len(intf))
    for i, item := range intf {
        floats[i] = item.(float64)
    }
    return floats
}

```

[Carving in Action](https://play.golang.org/p/eTFNLfgxSi)

It reduces lines of code, and eliminates some repeated logic, but it has to unmarshal the return value from `Carve` into the desired type, so, not as elegant as it'd otherwise be.

gengokit/gentesthelper/helper.go
```diff
@@ -0,0 +1,67 @@
...
+// examining how generated code differs from expected code.
+func DiffStrings(a, b string) string {
+	t := difflib.UnifiedDiff{
+		A:        difflib.SplitLines(a),
+		B:        difflib.SplitLines(b),
+		FromFile: "A",
+		ToFile:   "B",
+		Context:  5,
+	}
+	text, _ := difflib.GetUnifiedDiffString(t)
```
hasIan, line 65: Don't ignore errors.

gengokit/httptransport/httptransport.go
```diff
@@ -101,6 +111,24 @@ func NewBinding(i int, meth *deftree.ServiceMethod) *Binding {
 		nField.LowCamelName = LowCamelName(nField.Name)
 
 		nBinding.Fields = append(nBinding.Fields, &nField)
+
+		// Emit warnings for certain cases
+		if nField.IsBaseType == false {
```
hasIan, line 116: This could be expressed as `!nField.IsBaseType`.

gengokit/httptransport/httptransport.go
```diff
@@ -101,6 +111,24 @@ func NewBinding(i int, meth *deftree.ServiceMethod) *Binding {
...
+			wrn := []string{
+				"%s.%s is a custom type '%s', only base types and repeated base ",
+				"types are supported. As a result, the generated HTTP ",
+				"transport will fail to compile. Remove non-base types.",
+			}
+			log.Warnf(strings.Join(wrn, ""), meth.GetName(), nField.Name, nField.ProtobufType)
+		}
+		if field.Label == "LABEL_REPEATED" && nField.Location == "path" {
+			wrn := []string{
```
hasIan, line 125: Just use string concatenation instead of this slice of strings thing.

```
wrn := "%s.%s is a repeated field specified to be in the path. " +
    "Currently, repeated fields are not supported in the path " +
    "and may result in generated code which fails to compile."
```

Also, I would drop the word "currently" from that message. It sort of implies that we will support repeated fields as path variables. I don't think that will ever happen. If it does, the place to communicate it is in a roadmap or a tagged Github issue, not an error message in the executable.

gengokit/clientarggen/templated.go
```diff
@@ -0,0 +1,169 @@
...
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+//
+
+func CarveIntegerthirtytwo(inpt string) []int32 {
```
hasIan, line 11: I don't know about the "carve" metaphor; it's creative, but in software, that's a synonym for "clever," and cleverness is bad in software. Better would be a name makes plain what's going on just by reading it, like `ExtractInt32Slice` or `ConvertToInt32Slice`.

gengokit/clientarggen/client_argument_generator.go
```diff
@@ -43,10 +83,10 @@ type MethodArgs struct {
 //
 //     func Sum(ASum int64, BSum int64) (pb.SumRequest, error) {
 //              └────────────────────┘
-func (self *MethodArgs) FunctionArgs() string {
+func (m *MethodArgs) FunctionArgs() string {
 	tmp := []string{}
```
hasIan, line 87: If you know how big your end slice is going to be, declare it as such.

```
args := make([]string, 0, len(ma))
```

gengokit/clientarggen/client_argument_generator.go
```diff
@@ -23,15 +23,55 @@ import (
...
+	IsBaseType bool
+	// Repeated is true if this arg corresponds to a protobuf field which is
+	// given an identifier of "repeated", meaning it will represented in Go as
+	// a slice of it's type.
+	Repeated bool
 }
 
+// MethodArgs is a struct containing a slice of all the ClientArgs for this
+// Method.
 type MethodArgs struct {
```
hasIan, line 75: There seems to be no reason to make this a struct. Why not just

```
type MethodArgs []*ClientArg

func (ma *MethodArgs) FunctionArgs() string {
    args := make([]string, 0, len(ma))
    for _, a :=  range ma {
        append(args, fmt.Sprintf("%s %s", a.GoArg, a.GoType))
    }
    return args
}
```

gengokit/clientarggen/client_argument_generator.go
```diff
@@ -56,14 +96,30 @@ func (self *MethodArgs) FunctionArgs() string {
...
 	tmp := []string{}
-	for _, a := range self.Args {
-		tmp = append(tmp, createFlagConversion(*a))
+	for _, a := range m.Args {
+		tmp = append(tmp, a.GoArg)
 	}
 	return strings.Join(tmp, ", ")
 }
 
+// MarshalFlags returns the code for intantiating the GoArgs for this method
```
hasIan, line 107: "intantiating"

gengokit/clientarggen/strcarver.go
```diff
@@ -0,0 +1,138 @@
...
+//
+//     uint32
+//     uint64
+//     int32
+//     int64
+//     float32
+//     float64
+//     bool
+//     string
+func GenerateCarveFunc(m *ClientArg) string {
```
hasIan, line 29: You have "carve" functions for the primitive types actually implemented in this PR. Why do you need to generate a string form of a function? The contents of the template are near identical to the functions that are already implemented. Am I missing something, or is there redundancy here?

hasIan (changes requested):
Overall, looks very well thought out. Because I am not in your head and haven't thought about the conceptual problem you're addressing with these changes, and also because this is a large PR, this review was mainly cursory, as you can tell from the largely style-centric nature of the comments.

hasAdamr:
Could we use `json.Unmarshal` for the "Carving" functionally? It takes json arrays as strings and unmashals them into whatever slice type you give it. i.e.

```
        a := []int32{1, 2, 3, 4}
        s := `[1,2,3 ,4 ]`
        var b []int32
        err := json.Unmarshal([]byte(s), &b)
```

I did a few here: https://play.golang.org/p/xAO6luCtWI

After thinking about it, this is always how we can accept maps and nested messages. We just accept the field as json and unmarshal it.

gengokit/httptransport/httptransport.go
```diff
@@ -101,6 +111,24 @@ func NewBinding(i int, meth *deftree.ServiceMethod) *Binding {
...
+				"%s.%s is a repeated field specified to be in the path. ",
+				"Currently, repeated fields are not supported in the path ",
```
hasAdamr: Wait, won't maps and nested messages work if they are json in the body of a http request? It will just be unmarshaled.

gengokit/httptransport/httptransport.go
```diff
@@ -101,6 +111,24 @@ func NewBinding(i int, meth *deftree.ServiceMethod) *Binding {
 		nField.LowCamelName = LowCamelName(nField.Name)
 
 		nBinding.Fields = append(nBinding.Fields, &nField)
+
+		// Emit warnings for certain cases
+		if nField.IsBaseType == false {
+			wrn := []string{
+				"%s.%s is a custom type '%s', only base types and repeated base ",
+				"types are supported. As a result, the generated HTTP ",
```
hasAdamr, line 119: Wait won't maps and nested types work if there is json in the body of an http request? It will just be marshaled to the type.

gengokit/clientarggen/strcarver.go
```diff
@@ -0,0 +1,138 @@
...
+	parsefunc := ParseFunction(m.GoType)
+	typeconv := TypeConversion(m.GoType)
+
+	executor := struct {
+		ParseFunc string
+		TypeConv  string
+		GoArg     string
+		GoType    string
+		FlagType  string
```
hasAdamr, line 38: Would this ever be something other than string?

hasIan:
I like Adam's comment about simply using the JSON parser to interpret the repeated arguments instead of implementing our own special-purpose subset-of-JSON parser. This will allow support for repeated arguments with much less code. Less code is better.

hasAdamr:
I'm closing this now. No reason to leave it open, features are being worked on in other pull requests and this is diverging.

PR #63, opened 2016-09-29
Update template.go and remove $ go generate from travis

hasAdamr:
template.go has not been updated with the latest template files, so here they are.

`$ go generate` as a build step seems to be hurting us on this route as we sometimes forget to commit it even if templates has changed.

Because of this I have removed $ go generate from the .travis so that we test against what is committed rather than what is generated.


hasIan:
Make sense! LGTM. :shipit:

PR #64, opened 2016-09-29
Add support for repeated query parameters

hasLeland:
Adds support for repeated query parameters in the generated HTTP.

This functionality has been extracted from pull request #62 because that pull request is going to need some work, but this works and is helpful right now.

The comments that where left for that pull request have been addressed in this code.

### DIFF
--- a/gengokit/gentesthelper/helper.go
+++ b/gengokit/gentesthelper/helper.go
@@ -1,0 +1,64 @@
+package gentesthelper
+
+import (
+	"bytes"
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/printer"
+	"go/token"
+	"reflect"
+	"runtime"
+	"strings"
+
+	"github.com/pmezard/go-difflib/difflib"
+)
+
+// FuncSourceCode returns a string representing the source code of the function
+// provided to it.
+func FuncSourceCode(val interface{}) (string, error) {
+	ptr := reflect.ValueOf(val).Pointer()
+	fpath, _ := runtime.FuncForPC(ptr).FileLine(ptr)
+
+	funcName := runtime.FuncForPC(ptr).Name()
+	parts := strings.Split(funcName, ".")
+	funcName = parts[len(parts)-1]
+
+	// Parse the go file into the ast
+	fset := token.NewFileSet()
+	fileAst, err := parser.ParseFile(fset, fpath, nil, parser.ParseComments)
+	if err != nil {
+		return "", fmt.Errorf("ERROR: go parser couldn't parse file '%v'\n", fpath)
+	}
+
+	// Search ast for function declaration with name of function passed
+	var fAst *ast.FuncDecl
+	for _, decs := range fileAst.Decls {
+		if f, ok := decs.(*ast.FuncDecl); ok && f.Name.String() == funcName {
+			fAst = f
+			break
+		}
+	}
+	code := bytes.NewBuffer(nil)
+	err = printer.Fprint(code, fset, fAst)
+
+	if err != nil {
+		return "", fmt.Errorf("couldn't print code for func %q: %v\n", funcName, err)
+	}
+
+	return code.String(), nil
+}
+
+// DiffStrings returns the line differences of two strings. Useful for
+// examining how generated code differs from expected code.
+func DiffStrings(a, b string) string {
+	t := difflib.UnifiedDiff{
+		A:        difflib.SplitLines(a),
+		B:        difflib.SplitLines(b),
+		FromFile: "A",
+		ToFile:   "B",
+		Context:  5,
+	}
+	text, _ := difflib.GetUnifiedDiffString(t)
+	return text
+}
--- a/gengokit/httptransport/httptransport.go
+++ b/gengokit/httptransport/httptransport.go
@@ -11,6 +11,7 @@ import (
 	"text/template"
 	"unicode"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/TuneLab/go-truss/deftree"
 	"github.com/TuneLab/go-truss/gengokit/clientarggen"
 	gogen "github.com/golang/protobuf/protoc-gen-go/generator"
@@ -87,7 +88,7 @@ func NewBinding(i int, meth *deftree.ServiceMethod) *Binding {
 		var gt string
 		var ok bool
 		tmap := clientarggen.ProtoToGoTypeMap
-		if gt, ok = tmap[nField.ProtobufType]; !ok || field.Label == "LABEL_REPEATED" {
+		if gt, ok = tmap[nField.ProtobufType]; !ok {
 			gt = "string"
 			nField.IsBaseType = false
 		} else {
@@ -101,6 +102,24 @@ func NewBinding(i int, meth *deftree.ServiceMethod) *Binding {
 		nField.LowCamelName = LowCamelName(nField.Name)
 
 		nBinding.Fields = append(nBinding.Fields, &nField)
+
+		// Emit warnings for certain cases
+		if !nField.IsBaseType {
+			log.Warnf("%s.%s is a custom type '%s', only base types and repeated base "+
+				"types are supported. As a result, the generated HTTP "+
+				"transport will fail to compile. Remove non-base types.",
+				meth.GetName(),
+				nField.Name,
+				nField.ProtobufType)
+		}
+		if field.Label == "LABEL_REPEATED" && nField.Location == "path" {
+			log.Warnf(
+				"%s.%s is a repeated field specified to be in the path. "+
+					"Repeated fields are not supported in the path and may"+
+					"result in generated code which fails to compile.",
+				meth.GetName(),
+				nField.Name)
+		}
 	}
 	return &nBinding
 }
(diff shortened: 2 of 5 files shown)


gengokit/gentesthelper/helper.go
```diff
@@ -0,0 +1,67 @@
...
+	fset := token.NewFileSet()
+	fileAst, err := parser.ParseFile(fset, fpath, nil, parser.ParseComments)
+	if err != nil {
+		return "", fmt.Errorf("ERROR: go parser couldn't parse file '%v'\n", fpath)
+	}
+
+	// Search ast for function declaration with name of function passed
+	var fAst *ast.FuncDecl
+	for _, decs := range fileAst.Decls {
+		switch decs.(type) {
```
hasIan: This looks like it could be more succinctly replaced by a two-return-value type assertion. E.g.,

``` go
for _, decs := range fileAST.Decls {
    if f, ok := decs.(*ast.FuncDecl); ok && f.Name.String() == funcName {
        fAST = f
        break
    }
}
```

Added an early `break` out of the `range` if the function has been found here, too.

gengokit/gentesthelper/helper.go
```diff
@@ -0,0 +1,67 @@
...
+			if f.Name.String() == funcName {
+				fAst = f
+			}
+		}
+	}
+	code := bytes.NewBuffer(nil)
+	err = printer.Fprint(code, fset, fAst)
+
+	if err != nil {
+		return "", fmt.Errorf("couldn't print code for func '%v': %v\n", funcName, err)
```
hasIan: I suggest starting to get in the habit of using `%q` for arguments you know are strings that you want quoted. It's just easier.

gengokit/gentesthelper/helper.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	return code.String(), nil
```
hasIan: Is it actually important that this be returned as a `string`, or would `[]byte` actually be okay? In many ways, `[]byte` is more flexible than `string`. Early on in my Go tenure, a bit of advice to watch the conversions between `[]byte` and `string` was offered, that it was often unnecessary and done only because we are accustomed to working with strings, and that in high-throughput scenarios, it exacts a performance penalty. A couple years later, this seems to have been relevant advice. For example, the most effective way to work with things in Go is with `io.Reader` and `io.Writer`, interfaces to which byte slices lend themselves pretty nicely. Perhaps it is necessary in this case to have a string, for a reason that is not clear to me without additional context, but it is worth planting the seed to be mindful of conversions to `string`.

gengokit/httptransport/httptransport.go
```diff
@@ -6,17 +6,27 @@ import (
...
 	"github.com/TuneLab/go-truss/gengokit/clientarggen"
 	gogen "github.com/golang/protobuf/protoc-gen-go/generator"
 	"github.com/pkg/errors"
 )
 
+func init() {
+	log.SetLevel(log.InfoLevel)
+	log.SetOutput(os.Stderr)
+	log.SetFormatter(&log.TextFormatter{
+		ForceColors: true,
```
hasIan: `ForceColors` should probably be set accordingly depending on [whether the process is attached to a tty](https://rosettacode.org/wiki/Check_output_device_is_a_terminal#Go).

gengokit/httptransport/httptransport.go
```diff
@@ -101,6 +111,24 @@ func NewBinding(i int, meth *deftree.ServiceMethod) *Binding {
 		nField.LowCamelName = LowCamelName(nField.Name)
 
 		nBinding.Fields = append(nBinding.Fields, &nField)
+
+		// Emit warnings for certain cases
+		if nField.IsBaseType == false {
```
hasIan: I'm unaccustomed to seeing boolean checks with either `== true` or `== false`. It's redundant in the true case. I suppose it's a bit more explicit in the negated case, in that it's harder to miss than a `!` bang, but something still strikes me odd about explicit checks against `true` or `false`.

gengokit/httptransport/httptransport.go
```diff
@@ -101,6 +111,24 @@ func NewBinding(i int, meth *deftree.ServiceMethod) *Binding {
...
+			wrn := []string{
+				"%s.%s is a custom type '%s', only base types and repeated base ",
```
hasIan: Don't do this thing with a string slice that is `strings.Join`'d at runtime. Just use literal concatenation.

``` go
wrn := "%s.%s is a custom type '%s', only base types and repeated base " +
    "types are supported. As a result, the generated HTTP " +
    "transport will fail to compile. Remove non-base types."
```

gengokit/httptransport/httptransport.go
```diff
@@ -159,6 +187,43 @@ func (b *Binding) PathSections() []string {
 	return rv
 }
 
+func (f *Field) DecodeLogic() (string, error) {
```
hasIan: This function name, at a glance, leads me to believe that it is decoding something. However it is actually rendering something. This is complicated by the fact that it is used inside a template, whereupon it is treated like a property more than a function proper, in which case the noun-like naming makes sense.

Because of this, it seems okay to leave the naming suggesting that it's a property, although it smells a little bit to me. Perhaps more relevantly, I don't feel like "decode" is the right word here, as query string is not strictly an _encoding_ of something, like, for example, base64. I suggest a name like `QueryStringUnmarshaler`.

gengokit/httptransport/templates_test.go
```diff
@@ -201,7 +197,7 @@ func DecodeHTTPSumZeroRequest(_ context.Context, r *http.Request) (interface{},
 `
 	if got, want := str, desired; got != want {
```
hasIan, line 97: I'll probably keep objecting to this blind assignment of `got, want` in tests, _especially_ in the case where the values under test are already contained in variables. I'm not going to choose it as my hill to make a final stand on, but it mostly seems like extra verbosity to me.

hasIan (commented):
Nice-looking changes, though I've left some comments for your consideration. Nothing that I'd object so adamantly about so as to block a merge, but have a look and see what you think.

gengokit/httptransport/templates_test.go
hasAdamr: To me, where this is useful is when updating tests, sometimes the log message if forgotten and so this makes it so you only change the thing you are comparing against and the log updates too. i.e.

Rather than this:

```
    if got:= svc.GetName()  got != "ProtoService";  {
        t.Fatalf("\n`%v` was service name\n`%v` was wanted", got, "ProtoSerice")
    }
```

Do this:

```
    if got, want := svc.GetName(), "ProtoService"; got != want {
        t.Fatalf("\n`%v` was service name\n`%v` was wanted", got, want)
    }
```

I can see it being extra verbose though.

gengokit/gentesthelper/helper.go
hasLeland: Thank you for that advice, consider the seed planted!

In this particular case, we return `string` because when code is generated, it is returned as a string, and we want to do direct comparisons of the generated source code and the source of the functions used for testing.

gengokit/httptransport/httptransport.go
hasAdamr: I like how the []string is formatted better too, but Ian is probably right, not a good habit to get into.

gengokit/httptransport/httptransport.go
hasAdamr: Really this should be moved to main.go for truss, we can do the checks there and make these decisions. Probably should remove it for this pull request.

gengokit/httptransport/templates_test.go
hasIan: I can understand this, but this is a case I haven't really encountered; and the provided case ignores the "especially" clause in my objection above, when the "expected"/"want" value is already contained in a variable. It almost always is in my tests, as I'm typically using [table driven tests](https://github.com/golang/go/wiki/TableDrivenTests) and ranging over the slice of test cases.

If, for the sake of consistency, we wish to use this `got, want :=` pattern in our Truss tests, especially if they do not map nicely to being table driven, then I shan't object further. It's not terribly offensive and there's nothing actually _wrong_ with it; It's just something I haven't seen and wouldn't think about doing myself.

gengokit/httptransport/httptransport.go
ianfoo: I'd actually collapse it all down like so:

``` go
log.Warnf(
    "%s.%s is a custom type %q, only base types and repeated base " +
        "types are supported. As a result, the generated HTTP " +
        "transport will fail to compile. Remove non-base types.",
    meth.GetName(),
    nField.Name,
    nField.ProtobufType)
```

There's no need to declare a variable for it if all we're doing is using it the log statement below.

gengokit/httptransport/templates_test.go
hasLeland: Thanks for the call out on table driven tests Ian, I've seen them elsewhere and used them in many tests within truss, but I didn't know their name.

Also, as far as where we found this `got, want` pattern, it is largely drawn from [the tests for `grpc-gateway`, where they make heavy use of it.](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-grpc-gateway/descriptor/registry_test.go#L59-L81)

For right now, I'm going to leave this test logic how it is. But going forward, I will be trying to shift to more table driven tests.

gengokit/httptransport/httptransport.go
hasLeland: That makes sense, I'll remove these changes to the log settings.

gengokit/httptransport/httptransport.go
hasLeland: That's a good point, making the change.

gengokit/httptransport/httptransport.go
hasLeland: Done!

gengokit/httptransport/httptransport.go
hasLeland: The change is made!

hasLeland:
Alright, I believe I have addressed all items mentioned for this pull request. If the tests pass, then I'm going to merge the branch.

gengokit/httptransport/httptransport.go
```diff
@@ -101,6 +102,24 @@ func NewBinding(i int, meth *deftree.ServiceMethod) *Binding {
 		nField.LowCamelName = LowCamelName(nField.Name)
 
 		nBinding.Fields = append(nBinding.Fields, &nField)
+
+		// Emit warnings for certain cases
+		if !nField.IsBaseType {
+			log.Warnf("%s.%s is a custom type '%s', only base types and repeated base "+
+				"types are supported. As a result, the generated HTTP "+
+				"transport will fail to compile. Remove non-base types.",
+				meth.GetName(),
```
ianfoo, line 111: Is this how `gofmt` formatted it? Looks odd.

gengokit/httptransport/templates_test.go
ianfoo: Totally fine, sounds good.

gengokit/httptransport/httptransport.go
hasLeland: Yes, that is what `gofmt` formatted it to look like. I think it relates to the first string literal not being on the same line as the opening parenthesis. I'll probably change that at some future point, but it's at least passing `gofmt` now.

ianfoo: Oh, yes, I see, the first param is on the same line as the function call. I hadn't noticed that. When making multi-line calls, prefer placing the first parameter on the line after the function call. Just affirming what you said you intend to do.

PR #65, opened 2016-10-05
Update integration tests with http transport

hasAdamr:
Before the integration tests were only testing building the server and
cli client and testing for panics over grpc.

Now the following is tested:
- http transport, both the client library and manually set http requests
  - query, repeated query, and nested messages in body
- building cliclient and server
  - gprc and http transports
  - base, repeated, nested messages, and map type tests

The repeated, nested messages, and map type tests are currently disabled
until those features are implmented for the cliclient.

In addition a Makefile to the root directory has been added for building
truss and running all of the tests.


DEVELOPING.md
```diff
@@ -19,37 +21,58 @@ Then to build truss and it's two protoc plugins to your $GOPATH/bin directory:
...
-Before submitting a pull request always run tests that cover modified code. Also build truss and run truss's integration test. This can be done by
+Before submitting a pull request always run tests that cover modified code.
+Also build truss and run truss's integration test. This can be done by
 
-```
-$ cd $GOPATH/src/github.com/TuneLab/go-truss/truss
-$ make clean && make && make test
-# If the tests failed and you want to remove generated code
-$ make testclean
+```
```
hasLeland: There is a trailing space after this triple grave.

Additionally, this should mention that following these instructions will not run the integration tests, with instructions or directions to the instructions, on how to run the integration tests.

Makefile
```diff
@@ -0,0 +1,13 @@
+# Makefile for Truss.
+#
+# Build native Truss by default.
+default:
+	$(MAKE) -C truss
+
+# Run the go tests and the truss integration tests
+test:
+	go test -v ./...
```
hasLeland: This runs all the tess, including vendored tests, and that's pretty annoying. Can this be changed to maybe:

```
go test -v ./deftree/... ./gengokit/... ./gendoc/... ./truss/...
```

So that it only runs the tests for modules of truss?

truss/_integration-tests/README.md
```diff
@@ -0,0 +1,38 @@
+# integration-tests
+
+These tests run truss against definition files and tests if the generated code
+behaves. 
+
+# ./http
+
+The http directory contains http transport tests. They include tests against
+the generated http transport client libraries against the server and testing
```
hasLeland: I think this is intended to read:

> They include tests ~~against~~ **of** the generated http transport client libraries against the server**,** and testing hand-made HTTP requests against the server.

In general, remember to capitalize HTTP.

truss/_integration-tests/cli/test-service-definitions/1-basic/basic.proto
```diff
@@ -0,0 +1,58 @@
...
+  int64 D = 4;
+  uint32 E = 5;
+  uint64 F = 6;
+  sint32 G = 7;
+  sint64 H = 8;
+  fixed32 I = 9;
+  fixed64 J = 10;
+  sfixed32 K = 11;
+  bool L = 12;
+  string M = 13;
```
hasLeland, line 36: No complaints, I'm just really happy to have such thorough test cases to go work with!

hasLeland (commented):
In general, this looks like a huge improvement and I'm really excited to bring these changes in!

I do want to note that I've not reviewed much of the Go code in this PR, only verified that the test do run, and that they look really cool :100:.

DEVELOPING.md
hasAdamr: This will run the integration tests.

Makefile
hasAdamr: https://github.com/golang/go/issues/11659#issuecomment-122133295

It believe the default of testing the vendored libraries is a good idea. Whatever we have vendored is "our code" so running the tests the sake of completeness is great. 

Anyone developing truss should run this before making a pull request, and our ci should run it.

If you are working on a particular part of the code just run the tests for that code. Then check that everything is in order by running `make test` at the root.

Thoughts @hasIan?

truss/_integration-tests/README.md
hasAdamr: > > They include tests against of the generated **http** transport client libraries...
> 
> In general, remember to capitalize **HTTP**.

;p

Thanks, changed.

Makefile
hasIan: Initially I balked at the idea of running tests on vendored code, but after thinking about it for a second, the concept of "vendoring" is that it becomes _your_ code. So it makes sense to test it. Now, what is the likelihood of the vendored dependencies being broken? Pretty low, admittedly, for any dependency that's worth a hoot. However, the goal of software is--expressed in one manner--to eliminate uncertainty. In this regard, testing the code that we have adopted as our own seems reasonable.

We can always create an alternate make target (e.g., `test-nv`) to exclude vendored tests to help speed things up during development cycles, if that's the objection. Unit tests, properly categorized as unit tests, should execute very quickly though.

truss/_integration-tests/README.md
ianfoo: I find this whole sentence kind of awkward.

> They include tests of
the generated HTTP transport client libraries against the server as well as
testing hand made HTTP requests against the server.

Something like this seems clearer:

> The `http` directory contains tests for the generated HTTP handlers. Both the generated HTTP client and the standard Go `net/http` client are used to test the HTTP handlers.

I don't even know if that's necessary.

> The `http` directory contains tests for the generated HTTP handlers.

Substitute for "handlers" however is more accurate, if necessary.

hasIan: I find this whole sentence kind of awkward.

> They include tests of
> the generated HTTP transport client libraries against the server as well as
> testing hand made HTTP requests against the server.

Something like this seems clearer:

> The `http` directory contains tests for the generated HTTP handlers. Both the generated HTTP client and the standard Go `net/http` client are used to test the HTTP handlers.

I don't even know if that's necessary.

> The `http` directory contains tests for the generated HTTP handlers.

Substitute for "handlers" however is more accurate, if necessary.

DEVELOPING.md
hasLeland: Oh snap, you are correct. I was missing some test dependencies. Nevermind then.

hasAdamr:
I believe I have addressed all the comments now, but this needs additional review.

I have rebased ontop of Pull #66 which broke the test again repeated query params. 
It used to accept `?A=1&A=2` but now it only takes `?A=[1,2]`. We can add an issue for supporting the old style but this is fine.

Let me know what you think of the changes.

hasLeland:
Looks good to me!

PR #66, opened 2016-10-05
Add support for any msg structure in query and cli, excluding maps

hasLeland:
Adds support for handling any message structure (excluding messages with maps) in the query and in the cli.

This does not add support for map types, as map types are not correctly handled in `deftree`.


gengokit/httptransport/httptransport.go
```diff
@@ -104,15 +123,15 @@ func NewBinding(i int, meth *deftree.ServiceMethod) *Binding {
 		nBinding.Fields = append(nBinding.Fields, &nField)
 
 		// Emit warnings for certain cases
-		if !nField.IsBaseType {
-			log.Warnf("%s.%s is a custom type '%s', only base types and repeated base "+
-				"types are supported. As a result, the generated HTTP "+
-				"transport will fail to compile. Remove non-base types.",
+		if !nField.IsBaseType && nField.Location != "body" {
```
hasAdamr, line 131: > && nField.Location != "body"

<3

gengokit/clientarggen/client_argument_generator.go
```diff
@@ -73,16 +132,29 @@ type ClientServiceArgs struct {
...
-	for _, m := range self.MethArgs {
+	for _, m := range c.MethArgs {
 		for _, a := range m.Args {
 			tmp = append(tmp, a.FlagConvertFunc)
 		}
 	}
 	return strings.Join(tmp, "\n")
 }
 
+// AllCarveFuncs returns the code for each carve function for each repeated field.
```
hasAdamr: Clean up.

gengokit/clientarggen/client_argument_generator.go
```diff
@@ -109,75 +181,153 @@ func New(svc *deftree.ProtoService) *ClientServiceArgs {
...
+		newArg.Repeated = true
+	}
+	newArg.ProtbufType = field.Type.GetName()
+
+	newArg.FlagName = fmt.Sprintf("%s.%s", strings.ToLower(methName), strings.ToLower(field.GetName()))
+	newArg.FlagArg = fmt.Sprintf("flag%s%s", generatego.CamelCase(newArg.Name), generatego.CamelCase(methName))
+
+	if field.Type.Enum != nil {
+		newArg.Enum = true
+		//log.WithField("Arg", newArg.Name).Warnf("is an Enum!")
```
hasAdamr: Clean up.

gengokit/httptransport/httptransport.go
```diff
@@ -239,6 +259,28 @@ func createDecodeConvertFunc(f Field) string {
 	case strings.Contains(f.GoType, "string"):
 		fType = "%s := %s"
 	}
+	if !f.IsBaseType || f.Repeated {
```
hasAdamr, line 250: I feel this logic might be complex enough to go into it's own function. It is pretty hard to parse right here.

hasAdamr:
Question: Does this handle the type proto3 type `bytes`? If so, how so?

hasAdamr (changes requested):
We should avoid changing anything that causes us to generate code that does not build when for the same .proto file we used to generate code that does build.

i.e. The current `master` generates code that builds with `map<>`s in the .proto files, even if it did not function properly. 

Option 1: Add flag to not create cliclient. So no unrunnable code is created.
Option 2: Detect maps and ignore for client code
Option 3: Somehow detect that client will not build and either remove the client_main.go or _ the directory

I have done a sort hack pull request againt your branch for option 2. https://github.com/hasLeland/go-truss/pull/2

gengokit/httptransport/httptransport.go
hasLeland: well, I didn't put it in it's own function, but I have modified it extensively to be much more clear what it's doing, with comments explaining why. It may still benefit from being in a function, but let me know what you think of the improvements I've made so far.

hasAdamr (approved):

hasAdamr:
LGTM.

Map support is going to be implmented by @hasLeland by interpreting the .pb.go files that protoc-gen-go generates. This will simplify a lot of our pipeline for code generation.

Our `protoc ast` pipeline will be still be used for comment parsing to generate documentation at this time.

PR #68, opened 2016-10-10
Move to vendor with govendor instead of glide

tunezaq:
Proposal:
Switch over to govendor entirely. I switched my fork to be managed with
govendor which I find friendlier. I think it would be nice to just
switch truss over as well. If we wanna go that route then we can merge
this and follow up with a second PR to remove the glide files so we
don't get out of sync.


hasAdamr:
It seems the vendor.json file only contains thing that were not vendored before this pull request.

hasAdamr:
When I do govendor list I see that we are missing

```
  m golang.org/x/sys/unix                                     
```

And it seems that is used by `logrus`. And it does not have a vendor directory so I assume that `golang.org/x/sys/unix` is just not vendored yet.

hasAdamr:
When I do govendor list I see that we are missing
```
  m golang.org/x/sys/unix                                     
```

And it seems that is used by `logrus`. And it does not have a vendor directory so I assume that `golang.org/x/sys/unix` is just not vendored yet.

tunezaq:
Should be good now. See https://github.com/TuneLab/go-truss/pull/68/commits/bce1866f821f46c620a42b7f9204df92c1804235

hasAdamr:
Seems to be good to me.

Though two observations:
1. Should we just ignore solars like this, or actually vendor code that it will allow this to build on it? I am guessing we will just deal with this if someone makes an issue of it.
2. govendor seems to by default remove the tests of anything vendored. If one of these packages broke in some strange way on some machine then we would only see the results of our tests, and not get the granularity of our vendored code tests. This seems fine, just something that crossed my mind and seems to be a difference between glide and govendor.

tunezaq:
@hasAdamr per our conversation adding a means to do test vendor'd code in the CI should be good enough. Can you mark changes approved if that still sounds good to you?

solaris users should build with their local version of that packge and report any issues should they occur.

hasAdamr:
So how about on CI we can remove `test` from the ignore in `venor/vendor.json`

```
-       "ignore": "test, solaris",
+       "ignore": "solaris",
```

Then run

```
$ govendor sync
```

And govendor will grab all of the tests and their dependences. Then run `$ go test ./...` like normal.

We can use `sed` or something to remove `test` from `ignore`.

tunezaq:
@hasAdamr per our conversation, we don't need to be testing vendor'd code. Biggest reason being that our code should break horrifically should there be an issue.

PR #69, opened 2016-10-11
Split Docker images into build and run images

tunezaq:
Previously, there was a single Dockerfile defined that built Truss and
allowed users to run it. This yielded a tremendously large Docker image,
since much was downloaded and the entire Protobuf C++ implementation was
built inside it.

To reduce the size of the Docker image that most people will end up using,
one Dockerfile became two: one that builds an image with everything needed
to build Truss, and another that just runs Truss. When the build image is
run, the Truss source should be mounted into the container at
/go/src/github.com/TuneLab/go-truss, and the script that gets run by the
build image will build Truss into a build/ directory and also copy Truss'
runtime dependencies into build/.

Once this is complete, a runtime Truss image can be generated from the
contents of the build/ directory. The runtime image weighs in at about 25
MB, a fraction of the 1.6 GB combined image it replaces. The build image is
still around a gigabyte, but most users will not need to build Truss.

Add gosu and wrapper script to allow non-root-owned output

Writing files to a bind-mounted volume on the host that is running Docker
that aren't owned by root requires running as a user that maps to the UID
and GID that the volume is mounted as. This is achieved using a wrapper
script that launches the actual build script, to create user and group
entries, run the build script as the new user, and change ownership of
everything inside of the bind-mounted directory after the build script has
completed.

Update Go to latest versions, 1.6.3 and 1.7.1, and Protobuf to latest 3.0.2

Add missing period in comment

Move docker-specific Makefile into docker dir; default to native build

The Makefile in the root was added for Docker support and only dealt with
Docker-related targets. That's great for Docker, but it didn't make sense
to have the Makefile in the root of the project default to build Docker
images, since there's no requirement to even use Docker when using Truss.
As such, the Makefile in the project root has been reworked to default to
delegating to the Makefile in the truss directory, and the Docker-related
Makefile has been moved into the docker directory. The default target for
docker/Makefile will be available from the top-level Makefile.

ALSO: wrap docker run image to account prevent permission issues with its output


hasAdamr:
This does not seem to work with the Makefile.

```
Step 7 : COPY build/build-wrapper.sh /
lstat build/build-wrapper.sh: no such file or directory
```

If I change [this](https://github.com/zaquestion/go-truss/blob/ca94ddc969d1bf2a5225ea94221aa0428555d39b/docker/run/Dockerfile#L20) to this:

```
COPY docker/build/build-wrapper.sh /
COPY docker/run/run-truss.sh /
```

Then the copy works but I hit:

```
Step 9 : RUN -c /build-wrapper.sh $TRUSS_PATH /run-truss.sh
 ---> Running in f2ba3401ea2d
oci runtime error: exec: "-c": executable file not found in $PATH
Makefile:34: recipe for target 'build-docker-run' failed
```

Reproduce:

```
cd $GOPATH/src/github.com/TuneLab/go-truss/docker && make build-docker-build && make docker
```

Also if the makefile is changing, some name changes may be good too. Suggestion:

`build-docker-build` -> `build-docker-build-env`
`docker-build` -> `run-docker-build-env`
`build-docker-run` -> `build-docker-truss`
`docker-run` -> `run-docker-truss`

zaquestion:
Closed in favor of #138

PR #70, opened 2016-10-11
Add -pbout flag for .pb.go out directory; Large refactor for pipeline changes.

hasAdamr:
The -pbout flag has been added to specify a go package path to write the .pb.go files to. There is a cli test for this functionality.

Note that the imports in the handler files will not be updated if regenerated with a new -pbout flag. Also the -pbout flag must be used with the same package for each invoking of truss to keep the .pb.go files in that directory.

Useful information now prints when invoking truss without arguments, inspired by `cp` and `mv`.

In addition I have completely re-factored [`main.go`](https://github.com/hasAdamr/go-truss/blob/e765fa9d0e4889468a088fac5710aa19205d7383/truss/main.go) and [`protostage.go->protostuff.go`](https://github.com/hasAdamr/go-truss/blob/e765fa9d0e4889468a088fac5710aa19205d7383/truss/protostuff/protostuff.go). For `main.go` it will be hard to read the diff so I recommend just looking at the new file.

This refactoring was done to make it easier to add the -pbout flag, but also to have the code follow the flow for other cli programs such as `go build` and `go-bindata`.

Finally this refactoring was done in preparation for updating the pipeline with @hasLeland's .pb.go parser for a much more robust type system.


deftree/build_deftree.go
```diff
@@ -19,7 +19,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/TuneLab/go-truss/deftree/svcparse"
-	"github.com/TuneLab/go-truss/truss/protostage"
+	"github.com/TuneLab/go-truss/truss/protostuff"
```
zaquestion: 💩 smell. Surely there is a more descriptive name than `stuff`. Perhaps the package needs to be broken apart?

gengokit/gengokit_test.go
```diff
@@ -69,17 +69,13 @@ func TestNewTemplateExecutor(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	const goImportPath = "github.com/TuneLab/go-truss/gengokit"
+	const goImportPath = "github.com/TuneLab/go-truss/gengokit/general-service"
 
-	te, err := newTemplateExecutor(dt, goImportPath)
+	te, err := newTemplateExecutor(dt, goImportPath, goImportPath)
```
zaquestion: goPBImportPath?

hasAdamr: Same thing for these tests.

truss/protostuff/protostuff.go
```diff
@@ -0,0 +1,196 @@
...
+		return errors.Wrapf(err,
+			"protoc exec failed.\nprotoc output:\n\n%v\nprotoc arguments:\n\n%v\n\n",
+			string(outBytes), protocExec.Args)
+	}
+
+	return nil
+}
+
+// mkdir acts like $ mkdir -p path
+func mkdir(path string) error {
```
zaquestion: nitpick: I tend to be unfavorable torwards single line functions. I see that the filemode is being wrapped here so maybe that is case enough. Still I probably would have opted towards making the calls inline and using a package level constant. This makes it easier to see (without godef) what we are doing.

truss/truss/config.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"path/filepath"
+)
+
+// Config defines the inputs to a truss service generation
+type Config struct {
+	// The first path in $GOPATH
+	GOPATH string
+
+	// The path to where .pb.go files protoc-gen-go creates will be written
```
hasLeland: This comment could probably give a bit more context. For example:

> The path to where .pb.go files protoc-gen-go creates will be written. This defaults to the location of ServicePath, but could be any valid go path.

truss/protostuff/protostuff.go
```diff
@@ -0,0 +1,196 @@
+// Package protostuff provides an interface for interacting with protoc
+// requiring only paths to files on disk
+
+package protostuff
```
zaquestion: Glancing over this package contents and our other packages my proposal would be `protocgen` or just `protogen`

truss/truss/config.go
hasAdamr: I feel defaulting implies that this package will default it to that, which it could do, it just does not do that now. Only main.go knows this to be the case. I could add that comment though, I personally don't mind.

deftree/build_deftree.go
hasAdamr: It is the bag of code that deals with proto imports, execs protoc and does proto marshalling. Hence `protostuff`. I have called it `protocast`, `execprotoc`, and `protostage` but `protostuff` captures that this is code needed to interact with the protoc and be marshaled with the proto package.

I could break a function or two out of here and call it `execprotoc` if that is desired.

deftree/build_deftree.go
```diff
@@ -101,24 +101,21 @@ func NewFromString(def string) (Deftree, error) {
...
 	defer os.RemoveAll(protoDir)
 
-	err = ioutil.WriteFile(filepath.Join(protoDir, defFileName), []byte(def), 0666)
-	if err != nil {
-		return nil, errors.Wrap(err, "could not write proto definition to file")
-	}
+	defPath := filepath.Join(protoDir, defFileName)
 
-	err = protostage.Stage(protoDir)
+	err = ioutil.WriteFile(defPath, []byte(def), 0666)
```
zaquestion, line 106: On the 'all literals should be constants' debate I stray towards defining constants when you actually start using the value in multiple places. We don't appear to be here, which makes this OK in my book, but we should still leave a comment with some indication on why these permissions where chosen. Especially since they do not match the permission we use in protostuff/mkdir()

DEVELOPING.md
```diff
@@ -4,7 +4,7 @@
 
 1. Everything required to install `truss`
 2. go-bindata for compiling templates into binary `$ go get
-   github.com/jteeuwen/go-bindata`
+   github.com/jteeuwen/go-bindata/...`
```
hasIan: Favor block-style code markup when providing "type this at the command line" type instructions. They stand out more. E.g., `$ echo hello` vs.

``` bash
$ echo hello
```

deftree/build_deftree.go
hasIan: Agreed. "Stuff" makes "helper" and "util" seem useful and descriptive. What's wrong with `proto` or `pb` or even `protobuf`?

truss/protostuff/protostuff.go
zaquestion: Also the name of this function hides the fact that we are doing a `mkdir -p` under the hood. I defs think this should be moved to inline calls.

I realize its not new code, but its in your diff 😃

gengokit/gengokit.go
```diff
@@ -36,6 +36,8 @@ func init() {
 type templateExecutor struct {
 	// import path for the directory containing the definition .proto files
 	ImportPath string
+	// import path for .pb.go files containing service structs
```
hasIan, line 39: Use traditional sentence rules when writing comments that will be exported as documentation. I.e., capitalization, punctuation.

gengokit/gengokit_test.go
zaquestion: Sorry, shouldn't that addition be goPBImportPath?

gengokit/gengokit_test.go
```diff
@@ -172,7 +168,7 @@ func TestApplyTemplateFromPath(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	te, err := newTemplateExecutor(dt, goImportPath)
+	te, err := newTemplateExecutor(dt, goImportPath, goImportPath)
```
hasIan: This change hampers readability. I see five calls to `newTemplateExecutor`, all with the second parameter now duplicated to being the third parameter. As a reader, this raises questions in my mind that ought not to occur, like, "Why do I have to pass the same parameter twice?"

Will these parameters always be the same? If not, why are they the same here?

truss/truss/config.go
zaquestion: Could be worthwhile to add to the comment above the config struct itself instead to mention/point to where that these values have defaults exposed by the command arguments. I'm not sure theres a worthwhile comment that wouldn't go stale though. 

I'm inclined to agree with adam on the comment.

Separately is this a filepath or a go package? Name feels like the former, though we had been discussing the later. If it is the later we should probably rename to something `PBPackage`, `PBOutPackage`

truss/_integration-tests/cli/cli_test.go
```diff
@@ -342,16 +343,23 @@ func cleanTests(servicesDir string) {
...
-	}
+	os.RemoveAll(filepath.Join(defDir, "TEST-service"))
+	os.RemoveAll(filepath.Join(defDir, "bin"))
+	os.RemoveAll(filepath.Join(defDir, "pbout"))
+	mkdir(filepath.Join(defDir, "pbout"))
+}
+
+// mkdir acts like $ mkdir -p path
+func mkdir(path string) error {
+	// 0775 is the file mode that $ mkdir uses when creating a directoru
```
hasIan: "directoru"

truss/_integration-tests/cli/cli_test.go
```diff
@@ -342,16 +343,23 @@ func cleanTests(servicesDir string) {
...
+	err := os.MkdirAll(path, 0775)
```
hasIan: This function is effectively just `return os.Mkdir(path, 0775)`. I think I saw Zaq comment about this and although this is not a big deal, assigning `const filemodeMkdir = os.FileMode(0775)` and just calling `os.Mkdir(path, filemodeMkdir)` seems a little less overheady.

Also, `0775` is kind of a weird mode. Owner and group read/write/executable. Why not the more traditional `0755`?

gengokit/gengokit_test.go
hasAdamr:
```
func newTemplateExecutor(dt deftree.Deftree, goImportPath, goPBImportPath string) (*templateExecutor, error) {
```

This is because the template executor needs to know where the `goImportPath` of the service, is as well as where the `goPBImportPath` is in order to template the imports in the generated service. For these tests though the I'm just saying they will be in the same place as that is the case by default. I could add a comment about each about this.

gengokit/gengokit_test.go
hasAdamr:
```
func newTemplateExecutor(dt deftree.Deftree, goImportPath, goPBImportPath string) (*templateExecutor, error) {
```

This is because the template executor needs to know where the `goImportPath` of the service is, as well as where the `goPBImportPath` is in order to template the imports in the generated service. For these tests though the I'm just saying they will be in the same place as that is the case by default. I could add a comment about each about this.

gengokit/gengokit_test.go
zaquestion: I actually thought that was a typo up until this comment. I would propose just declaring a second constant.

deftree/build_deftree.go
hasAdamr: Well I am importing a packaged named `proto` in `protostuff` but if you think it would be fine to just call it `proto` I would be down for that.

truss/main.go
```diff
@@ -6,133 +6,209 @@ import (
...
 )
 
-func main() {
+var (
+	pbPathFlag = flag.String("pbout", "", "The go package path where the protoc-gen-go .pb.go structs will be written.")
+)
+
+func init() {
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, "Usage: %s [OPTION]... [*.proto]...\n", os.Args[0])
```
hasIan: Recommend using `path.Base(os.Args[0])` to clean up the invocation name.

truss/main.go
```diff
@@ -6,133 +6,209 @@ import (
...
+
+func main() {
+	cfg, err := parseInput()
+	exitIfError(errors.Wrap(err, "could not parse input"))
+
+	dt, err := parseServiceDefinition(cfg.DefPaths)
+	exitIfError(errors.Wrap(err, "unable to parse input definition proto files"))
+
+	err = updateConfigWithService(cfg, dt)
+	exitIfError(errors.Wrap(err, "unable to read system based on service definition"))
```
hasIan: What does "read system" mean?

truss/protostuff/protostuff.go
hasAdamr: `protogen` could be good. Many of these functions parse things rather than generate things. @hasIan  @hasLeland, `protogen`?

truss/main.go
```diff
@@ -6,133 +6,209 @@ import (
...
+
+// parseInput constructs a *truss.Config with all values needed to parse
+// service definition files.
+func parseInput() (*truss.Config, error) {
+	var cfg truss.Config
+
+	// GOPATH
+	goPaths := filepath.SplitList(os.Getenv("GOPATH"))
+	if goPaths == nil {
+		return nil, errors.New("$GOPATH not set")
```
hasIan: Minor: the name of the variable is `GOPATH`, not `$GOPATH`, which represents the value.

truss/main.go
hasAdamr: read filesystem?

This error will only come about if for some reason truss is unable to read previously generator files. I had trouble writing this error.

zaquestion: 👍 to making it filesystem and maybe add as a comment

```
This error will only come about if for some reason truss is unable to read previously generator files. I had trouble writing this error.
```

truss/main.go
```diff
@@ -6,133 +6,209 @@ import (
...
+	var err error
 	rawDefinitionPaths := flag.Args()
+	cfg.DefPaths, err = cleanProtofilePath(rawDefinitionPaths)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to parse input arguments")
+	}
 
-	protoDir, definitionFiles, err := cleanProtofilePath(rawDefinitionPaths)
+	// PBGoPath
+	if *pbPathFlag != "" {
```
hasIan: I'm debating inverting this to keep the code from forming an `if`-triangle here. Like so:

``` go
if *pbPathFlag == "" {
    return &cfg, nil
}
cfg.PBGoPath = filepath.Join(cfg.GOPATH, "src", *pbPathFlag)
if !fileExists(cfg.PBGoPath) {
    return nil, errors.Errorf(".pb.go output package directory does not exist: %q", cfg.PBGoPath)
}
return &cfg, nil
```

I'm not crazy about the same return being exactly repeated, but it allows for optimistic exit in the case that `-pbout` hasn't been specified, rather than guarding positive-path logic inside an `if`.

Also, what's the `cfg.GOPATH`? All-caps variable names are not idiomatic Go, even though we see them used often by former Java programmers. 😉

truss/main.go
```diff
@@ -6,133 +6,209 @@ import (
...
-		exitIfError(errors.Wrapf(err, "could not open %v", protoF))
+	svcFile, err := protostuff.ServiceFile(protocOut, filepath.Dir(definitionPaths[0]))
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to find service definition file")
+	}
 
-		files = append(files, protoF)
+	dt, err := deftree.New(protocOut, svcFile)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to construct service defintion")
```
hasIan: "defintion"

truss/main.go
hasAdamr: I personally like the inversion, flattens things out even if the return is repeated. 

I chose all caps for `GOPATH` because that is what it is called in the go build context struct: https://github.com/golang/go/blob/master/src/go/build/build.go#L34

truss/main.go
```diff
@@ -145,40 +221,40 @@ func mkdir(path string) error {
...
+func readPreviousGeneration(serviceDir string) ([]truss.NamedReadWriter, error) {
+	if fileExists(serviceDir) != true {
 		return nil, nil
 	}
 
 	var files []truss.NamedReadWriter
 	sfs := simpleFileConstructor{
-		protoDir: protoDir,
-		files:    files,
+		dirPath: filepath.Dir(serviceDir),
```
hasIan: Super minor, but `dirPath` feels like naming something `filefile`.

truss/protostuff/protostuff.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	genGoCode := "--go_out=Mgoogle/api/annotations.proto=" +
+		goImportPath + "/third_party/googleapis/google/api," +
+		"plugins=grpc:" +
+		outDir
+
+	_, err = exec.LookPath("protoc-gen-go")
+	if err != nil {
+		return errors.Wrap(err, "protoc-gen-go not exist in $PATH")
```
hasIan: Maybe add "does" in here to clean up the robotic-sounding "not exist"? Or perhaps better and more human-targeted, "cannot find progoc-gen-go in PATH."

truss/protostuff/protostuff.go
```diff
@@ -0,0 +1,196 @@
...
+
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not open service file: %v\n in path: %v",
+			protoFileDir, svcFileName)
+	}
+
+	return svc, nil
+}
+
+// getProtocOutput calls exec's $ protoc with the passed protofiles and the
```
hasIan: > calls exec's $ protoc

Either I'm confused by this, or it should just be "executes protoc," or similar.

truss/protostuff/protostuff.go
```diff
@@ -0,0 +1,196 @@
...
+		return nil, errors.Wrap(err, "protoc failed")
+	}
+
+	fileInfo, err := ioutil.ReadDir(protocOutDir)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not read directory: %v", protocOutDir)
+	}
+
+	for _, f := range fileInfo {
+		if !f.IsDir() {
```
hasIan: Can flatten this by inverting the `f.IsDir` call. E.g.,

``` go
if f.IsDir() {
    continue
}
// ...
```

truss/truss/config.go
hasAdamr: In this config it is the full path. You specify it on the command line with the go package name though. 

It is this way because I was passing this value to protoc to inform it where to write the .pb.go files. 

Though thinking about it because I already have the go path I can just `filepath.Join(cfg.GOPATH, "src", cfg.PBPackage)` to get the full path for protoc.

truss/main.go
hasAdamr: The `directory` and `path` names have always confused me.

> A **directory** is a "folder", a place where you can put files or other directories (and special files, devices, symlinks...). It is a **container for filesystem objects**. A **path** is a **string** that specify how to reach a filesystem object...

So `dirPath` is saying the path to the directory like filepath is saying the same thing for a file. 

Because of this I was really tempted to change `serviceDir` and such to `serviceDirPath` but that is so verbose, even if it is technically correct.

Just calling it `path` though seemed bad because at the time I was importing `path`. 

Do you have an alternative name suggestion?

truss/_integration-tests/cli/cli_test.go
hasAdamr: I chose `0775` because when I make a directory on linux with the command `mkidr` it has the permissions of `0775`. This is also why I wrapped it in a function, but I guess that is system specific. Why is `0755` more traditional? Wouldn't you want other group owners to be able to write as well?

I will pull this out into a const. and all it with os.MkdirAll though. Good suggestion.

truss/truss/truss.go
```diff
@@ -1,5 +1,6 @@
 // Package truss contains the relative file tree data structure that represents
 // the paths and contents of generated files
+
```
hasIan: Do not put a blank line between the package documentation and the `package` line. Godoc will not pick up the comments as documentation. This applies to anything that's intended to be documentation: types, consts, vars, and functions.

hasIan: Additionally, and not specific to this reivew, we must be mindful of how we're writing our documentation comments. For example, here's what I get when I ask for documentation on the `truss` package:

```
➜  godoc github.com/TuneLab/go-truss/truss/truss
use 'godoc cmd/github.com/TuneLab/go-truss/truss/truss' for documentation on the github.com/TuneLab/go-truss/truss/truss command

PACKAGE DOCUMENTATION

package truss
    import "github.com/TuneLab/go-truss/truss/truss"


TYPES

type NamedReadWriter interface {
    io.ReadWriter
    Name() string
}
    NamedReadWriter represents a file name and that file's content Name() is
    a path relative to the directory containing the .proto files Name()
    should start with "service/" for all generated and read in files os.file
    fulfills the NamedReadWriter interface
```

That `NamedReadWriter` is a glob of text that muddies things, not clarifies things.

deftree/build_deftree.go
hasIan: Nah, I didn't look that closely. Avoid intentional name collisions. :) Just something that describes what it does better than "stuff."

gengokit/gengokit_test.go
hasIan: A clearer interface here, I think, would be passing `""` as the third parameter if it's going to be the same. If it's just for testing, though, then it's not a big deal.

truss/protostuff/protostuff.go
hasIan: If it's doing more parsing than generation, then wouldn't `protoparse` be more apropos? Am I misunderstanding? Ideally, parsing and rendering should be separated so the packages could be named explicitly after what they're actually doing.

gengokit/gengokit_test.go
hasIan: Seeing this declaration, I'm further nudged toward the idea that passing `""` as `goPBImportPath` should be interpreted by the function as, "Just use the `goImportPath` as the protobuf import path," as long as it is documented in the function comments. Passing a zero value to allow a default value to be used is something established in the Go community. Thoughts?

truss/_integration-tests/cli/cli_test.go
hasIan: I've never seen 0775 as a default, personally. 0755 for directories and executables, and 0644 for regular files, is what I've seen nearly everywhere. What you get depends on your umask value (often 022). But typically, no, I would not want anyone other than myself to write the files that I create. Like, I'm writing code for my awesome new product, and that joker Steve comes along and overwrites all the amazing algorithms with ASCII lolcats and deleted the `.git` directory. Dammit, Steve, you've gone too far!

truss/main.go
hasIan: "Unable to read filesystem based on service definition" is not any clearer to me. I'd say it should reflect what's happening immediately before, but unfortunately, "update config with service" isn't much clearer to me either.

hasIan: Per Zaq's comment, if that's accurate, the error should be something like, "unable to read previously-generated output," although ideally something that indicates why that's a problem. Our audience is humans who don't necessarily care how Truss works, just that it does.

truss/main.go
hasIan: Fair enough on `GOPATH`. I've come around to liking the flattened structure, too, in spite of the repeated `return` line. 👍

truss/main.go
hasIan: That definition makes sense; I can't argue with it. And by it, `dirPath` makes sense. It just sounds weird. I understand not calling it `path`. I probably would have just called it `dir`. But you've provided an adequate defense, if you want to keep it `dirPath`. It just makes my eyebrows twist a little to see it. :)

truss/main.go
hasAdamr: Haha I love your message because this error wraps this error: [unable to read previously generated files](https://github.com/hasAdamr/go-truss/blob/e765fa9d0e4889468a088fac5710aa19205d7383/truss/main.go#L125)

Should I wrap it again? This was a place I got stuck.

hasIan: Haha, I probably should have looked more closely. But in this case, if you're going to wrap it, it should add value, and given the current error message, I don't think it does. "unable to read previously generated files" is clear enough, but if I got that message while running, there's a good chance I don't know why that's important, or what I need to do to fix it.

Looking at this example specifically, though, the error is only going to be returned if the next deeper level in the stack has a problem walking the directory tree. That's probably not going to happen, and if it does, I have three levels of errors without even wrapping it again:
1. unable to read previously generated files
2. could not to fully walk directory %s
3. error from `filepath.Walk`

I'd argue that this likely presents enough of a picture here without having to wrap again. We're not _obligated_ to wrap errors in every level.

As an aside, we should keep in mind the language we're using in UI messages and be consistent. E.g., "could not" vs "unable to." Minor, but consistency in the user interface is a good thing.

truss/truss/config.go
zaquestion: I like that idea. Although what happens when the GOPATH is composed of multiple paths?

gengokit/gengokit_test.go
zaquestion: No more commenting here dupe comment thread below.

truss/truss/config.go
hasIan: Truss needs to accommodate path-lists in `GOPATH`, if it doesn't already.

``` go
cfg.GOPATH = strings.Split(os.Getenv("GOPATH"), string(os.PathListSeparator))[0]
```

gengokit/gengokit_test.go
zaquestion: @hasIan I wanna say that I'm torn, but I can't actually seem to form a concrete idea about why. I think its probably fine to use the zero value to default the path. So long as the function has a comment stating the behavior (for godoc)

hasIan: I feel like I shouldn't _have to_ repeat myself when invoking the function if I don't want to take advantage of the optional location to place the `.pb.go` files, and it's already a convention in Go codebases. However, it _is_ a little bit of "magic." You have to know what `""` implies if you're reading the code, whereas the same reference repeated, you know what's going to be used for each value without having to know what `""` means. I guess clearest would be to have two functions, but that's clearly overkill, as, pretty much, is this whole discussion. It's not a big deal in this instance, but it's also important in the longer term that we align around an approach.

truss/truss/config.go
adamryman: It always picks the first path in the gopath like the go tooling.

```
    // GOPATH
    goPaths := filepath.SplitList(os.Getenv("GOPATH"))
    if goPaths == nil {
        return nil, errors.New("$GOPATH not set")
    }
    cfg.GOPATH = goPaths[0]
```

https://github.com/hasAdamr/go-truss/blob/e765fa9d0e4889468a088fac5710aa19205d7383/truss/main.go#L67

hasAdamr:
I belive I have addressed all comments in this PR. If anyone would like to take a second look after my last commit please do so.

hasAdamr:
I belive I have addressed all comments in this PR. If anyone would like to take a second look after my last commit please do so.

truss/truss/truss.go
```diff
@@ -9,11 +8,11 @@ import (
...
 
 // NamedReadWriter represents a file name and that file's content
-// Name() is a path relative to the directory containing the .proto files
-// Name() should start with "service/" for all generated and read in files
-// os.file fulfills the NamedReadWriter interface
 type NamedReadWriter interface {
 	io.ReadWriter
+	// Name() is a path relative to the directory containing the .proto
+	// files Name() should start with "service/" for all generated and read
+	// in files os.file fulfills the NamedReadWriter interface
```
hasLeland: This comment is still really unclear. It looks like it was originally intended to be three separate sentences:
1. `Name()` is a path relative to the directory containing .proto files.
2. `Name()` should start with "service/" for all generated and read in files.
3. `os.files` fulfills the `NamedReadWriter` interface.

With that in mind, I think it reads more clearly when broken up as definitely seperate sentences:

```
// Name() is a path relative to the directory containing the .proto
// files.  Name() should start with "service/" for all files which have
// been generated and read in. Note that os.file fulfills the
// NamedReadWriter interface.
```

hasLeland: Other than this documentation change, however, this...

## Looks good to me!

PR #71, opened 2016-10-12
New service definition implementation -- 1/5 the size of Deftree!

hasLeland:
Before we get to the "what" of all this, here's some quick notes:
1. I'm pretty sure this needs more documentation. Please do clone this down and look at it using the godoc web interface or CLI and let me know if there're things which aren't clear from the existing docs.
2. What do you think about naming it `svcdef`? Should that be changed? What do you think would be the best name?

---
# What?

In short, `deftree` is a big mess that's not good at what we use it for, so we're gonna replace it with this new thing called `svcdef`.
# Why?

`deftree` is intended to hold all the info we may need to generate a service. However, `deftree` exclusively derives it's information from the parsing of the input proto files. This sounds fine in priciple, but when generating code, we have to generate code which co-operates with the go code generated by `protoc-gen-go`. Now, you may notice a problem here: how can we generate code to interoperate with the code from `protoc-gen-go` if we haven't directly inspected the output code from `protoc-gen-go`?

The answer till now has been: by making some (not always sound) assumptions. Since the code generated by `protoc-gen-go` is also based on the same protobuf definition files we are provided, it is possible _in theory_ to infer the state of the code generated by `protoc-gen-go` and generate our own code which interoperates with our correctly-infered code. But our inferences aren't always correct, and even if they are they frequently don't handle edge cases in the code we must interoperate with.

If this sounds overly complicated and error prone, that's because it has been.

In short, **`deftree` only indirectly gives us the information we need to generate working and reliable code, which is why `svcdef` has been written.**
# How?

How does `svcdef` suceed where `deftree` failed? By getting it's information from the go files we need to interoperate with. Instead of deriving it's information from protobuf files, `svcdef` derives its info from the go files created by `protoc-gen-go` (and by reading the protobuf files for their http annotations). By sourcing it's information directly from the go files, `svcdef` provides much more accurate information that doesn't have to be massaged to generate correct code in the way the output of `deftree` does. Additionally, because the golang `go/ast` package for parsing go files is :+1::100:amazing, we can extract this more accurate and useful information with **80% less code** than `deftree`!
# Does this break anything?

As of right now, no! But that's because this is only the implementation of `svcdef`, this PR does not include the changes to the rest of `truss` to use `svcdef` instead of `deftree`. That change is going to come later. The first goal is to get eyes on this code before we spend more time migrating things over.


svcdef/consolidate_http.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/pkg/errors"
+)
+
+// ConsolidateHTTP accepts a SvcDef and the io.Readers for the proto files
+// comprising the definition. It modifies the SvcDef so that HTTPBindings and
+// their associated HTTPParamters are added to each ServiceMethod. After this,
+// each `HTTPBinding` will have a populated list of all the http parameters
+// that that binding requires, where that parameter should be located, and the
+// type of each parameter.
+func ConsolidateHTTP(sd *Svcdef, protoFiles []io.Reader) error {
```
hasLeland: `ConsolidateHTTP` may not need to be exported, as the functionality it provides is generally only for use during the creation of a new `Svcdef`. So this should probably be made unexported. Thoughts from anyone else?

svcdef/consolidate_http.go
```diff
@@ -0,0 +1,122 @@
...
+			} else {
+				return errors.Wrap(err, "error while parsing http options for the service definition")
+			}
+		}
+		assembleHTTPParams(sd.Service, protosvc)
+	}
+	return nil
+}
+
+func assembleHTTPParams(svc *Service, httpsvc *svcparse.Service) error {
```
hasLeland, line 41: `assembleHTTPParams` is named somewhat vaguely, and would probably benefit from a doc-comment explaining its purpose, as well as a few comments providing naration of its actions.

svcdef/resolvetypes.go
```diff
@@ -0,0 +1,61 @@
+package svcdef
+
+// After the initial creation of Svcdef, each FieldType will have it's name
+// set to the name of the type that it represents, but it won't have a pointer
+// to the full instance of the type with that name. Type resolution is the
+// process of taking the name of each FieldType, then searching for the
+// enum/message with that name and setting the respective *Message/*Enum
+// property of our FieldType to point to that enum/message.
```
hasLeland: This comment is currently just "floating" in the middle of nowhere, and isn't tied to this package or any functions within this file. This explanation should probably be moved into the doc comment for the `resolveTypes(*Svcdef)` function below.

svcdef/resolvetypes.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	if sd.Service != nil {
+		for _, m := range sd.Service.Methods {
+			setType(m.RequestType, tmap)
+			setType(m.ResponseType, tmap)
+		}
+	}
+}
+
+func setType(f *FieldType, tmap map[string]typeBox) {
```
hasLeland, line 42: The purpose of `setType` isn't clear. There should be a doc-comment for this function providing an explanation of what it does and why it's here.

svcdef/resolvetypes.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	switch {
+	case entry.Enum != nil:
+		f.Enum = entry.Enum
+	case entry.Message != nil:
+		f.Message = entry.Message
+	}
+}
+
+// newTypeMap returns a map from
```
hasLeland: ...from what!?

The doc comment for newTypeMap is incomplete. Finish adding a complete explanation.

svcdef/svcdef.go
```diff
@@ -0,0 +1,347 @@
+package svcdef
+
+// consider renaming to census, pbregistry, or just to "essence"
```
hasLeland: This comment was intended as a note for myself on potential names for this package, and needs to be removed.

svcdef/svcdef.go
```diff
@@ -0,0 +1,347 @@
...
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+type Svcdef struct {
```
hasLeland, line 30: `Svcdef` lacks an explanatory doc-comment. In fact, here's a list of the exported types in this file which need documentation added:

```
1. Svcdef
2. Message
3. Enum
4. Map
5. Service
6. ServiceMethod
7. Field
8. FieldType
```

svcdef/svcdef.go
```diff
@@ -0,0 +1,347 @@
...
+	resolveTypes(&rv)
+	err := ConsolidateHTTP(&rv, protoFiles)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to consolidate HTTP")
+	}
+
+	return &rv, nil
+}
+
+// NewEnum returns a new Enum struct derived from an *ast.TypeSpec
```
hasAdamr: This comment is the same as reading the function signature. We should avoid having these types of comments unless they say something that is non obvious from the function signature as the documentation contains these. 

2016-10-13-142056_562x148_scrot

svcdef/svcdef.go
```diff
@@ -0,0 +1,347 @@
...
+	}
+	rv.ResponseType, err = makeFieldType(rs)
+	if err != nil {
+		return nil, errors.Wrapf(err, "ResponseType creation of service method %q failed", rv.Name)
+	}
+
+	return rv, nil
+}
+
+// NewField returns a Field struct with information distilled from an
```
hasAdamr: Another comment that is not helpful in the docs, possibly bring the chart up from below. Or remove or give context of why an error might be returned.

2016-10-13-142114_631x138_scrot

svcdef/svcdef.go
```diff
@@ -0,0 +1,347 @@
...
+	// ArrayType is True if this FieldType represents a slice of a type.
+	ArrayType bool
+}
+
+// HTTPBinding represents one of potentially several bindings from a gRPC
+// service method to a particuar HTTP path/verb.
+type HTTPBinding struct {
+	Verb string
+	Path string
+	// There is one HTTPParamter for each of the parent service methods Fields.
```
hasAdamr: Below you use `Service.RequestType.Fields` and here you say `service method Fields`. The RequestType information is helpful, keep that but make the naming/caps/dots consistent thoughtout documentation.

hasLeland:
Alright, I believe I've adressed all the items in this code review.

Anyway, @hasIan and @tunezaq I'm eager to hear what you two think!

tunezaq:
Some review and general brain dump interleaved, enjoy :)

I'm noticing a log a of large comment blocks, in some cases rivaling the function body. Which I love compared to the alternative. Still I think we should rely a bit more of go's integrity and aim to not repeat what can be read in code. For comments some goods rules of thumb are:
stray from the bare `what`
`why` this way and not just `X`, aka "what are the needs?"
`where` are side effects, external impact, "what is this for?"

The reasons for this are that the `what` is almost always in the function or block itself. Try to create function [names] that imply the underly code. Otherwise inlining may be better suited, provided you aren't senselessly duplicating code.

The `why` can stray across several functions, packages, services, filesystem, hardware, product, language, preferences, mental state, etc... Its one of the hardest to explain and one of the least obvious in any one location.

`where` can be figured out pretty easy in a lot of cases, but tips offs are nice. You can think of this section as "what is this for?" so long as you giving semi decent direction so the thing.

svcdef/resolvetypes.go
```diff
@@ -0,0 +1,71 @@
...
+	switch {
+	case entry.Enum != nil:
+		f.Enum = entry.Enum
+	case entry.Message != nil:
+		f.Message = entry.Message
+	}
+}
+
+// newTypeMap returns a map of names of types (Messages/Enums) to typeBoxes,
+// which contain a reference to that Message/Enum.
```
tunezaq: Consider

```
newTypeMap returns a map of FieldType Names to correlated typeBoxes
```

svcdef/resolvetypes.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+}
+
+// setType uses a map of type-names to types to resolve the type of a
+// particular FieldType. In the first 'pass' of svcdef creation, messages will
+// be defined, but only the names of the types of their fields are known;
+// fields won't contain a reference to their type. But each fieldtype should
+// also contain a reference to the definition of that type. The creation of
+// that reference, from *FieldType f to the definition of an Enum, Message, or
+// Map, is what setType does.
```
tunezaq: Consider

```
setType unpacks typeBox value into corresponding FieldType
```

svcdef/resolvetypes.go
```diff
@@ -0,0 +1,71 @@
...
+// represents, but it won't have a pointer to the full instance of the type
+// with that name. Type resolution is the process of taking the name of each
+// FieldType, then searching for the enum/message with that name and setting
+// the respective *Message/*Enum property of our FieldType to point to that
+// enum/message.
+func resolveTypes(sd *Svcdef) {
+	tmap := newTypeMap(sd)
+	for _, m := range sd.Messages {
+		for _, f := range m.Fields {
+			setType(f.Type, tmap)
```
tunezaq, line 30: I believe this to be a bug. IIRC the `f` here is a copy used while ranging over `m.Fields`. This can easily be checked by grabbing the index and comparing their addresses.
https://golang.org/ref/spec#For_statements

```
Range expression                          1st value          2nd value

array or slice  a  [n]E, *[n]E, or []E    index    i  int    a[i]       E
string          s  string type            index    i  int    see below  rune
map             m  map[K]V                key      k  K      m[k]       V
channel         c  chan E, <-chan E       element  e  E
```

see second value `a[i]` vs `&a[i]`

svcdef/svcdef_test.go
```diff
@@ -0,0 +1,131 @@
...
+				t.Errorf("Field %q on message %q has type which differs from the typemap type of the same name, got %p, want %p", selectedf.Name, msg.Name, selectedf.Type, ftype)
+			}
+		}
+	}
+}
+
+// Test that type resolution of map values functions correctly. So if a message
+// has a map field, and that map field has values that are some other message
+// type, then the type of the key will be correct.
+func TestMapTypeResolution(t *testing.T) {
```
tunezaq: Only tests newTypeMap

svcdef/resolvetypes.go
tunezaq: You test for TypeResolution appears to only assert on the values created after newTypeMap. Testname should be updated to reflect or function changed to test this function

svcdef/resolvetypes.go
```diff
@@ -0,0 +1,71 @@
...
+// child structs, in place. After the initial creation of Svcdef, each
+// FieldType will have it's name set to the name of the type that it
+// represents, but it won't have a pointer to the full instance of the type
+// with that name. Type resolution is the process of taking the name of each
+// FieldType, then searching for the enum/message with that name and setting
+// the respective *Message/*Enum property of our FieldType to point to that
+// enum/message.
+func resolveTypes(sd *Svcdef) {
+	tmap := newTypeMap(sd)
+	for _, m := range sd.Messages {
```
tunezaq: should `sd.Enums` be iterated over here? Likely no, I believe Enums still have to be used inside a Message and are exposed through `m.Fields`, but this is not obvious if you don't know protobuf or simply might not be in mind. Consider adding a short comment to remind the reader.

svcdef/resolvetypes.go
```diff
@@ -0,0 +1,71 @@
+package svcdef
+
+// typeBox is used where we know that either a Message or an Enum will need to
+// be returned, but not both, and not neither. If one field is nil, the other
+// is expected to be a valid pointer.
```
tunezaq: Consider

```
Holds either a Message or an Enum; used only in resolveTypes() to 
associate FieldTypes with their underlying data.
```

svcdef/resolvetypes.go
```diff
@@ -0,0 +1,71 @@
...
+type typeBox struct {
```
tunezaq: Consider placing next to newTypeMap, nothing says types have to go at the top.

svcdef/resolvetypes.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// resolveTypes accepts a pointer to a Svcdef and modifies that Svcdef, and its
+// child structs, in place. After the initial creation of Svcdef, each
+// FieldType will have it's name set to the name of the type that it
+// represents, but it won't have a pointer to the full instance of the type
+// with that name. Type resolution is the process of taking the name of each
+// FieldType, then searching for the enum/message with that name and setting
+// the respective *Message/*Enum property of our FieldType to point to that
+// enum/message.
```
tunezaq: Consider

```
Set underlying types for all of our service methods Fields, RequestTypes, and responseTypes
```

tunezaq (commented):
Heres a thorough review of resolvetypes.go. I would like to review the rest of the code as well, but not in this sitting.

tunezaq:
test

svcdef/svcdef.go
```diff
@@ -0,0 +1,354 @@
...
+			keyFollower(ex.X)
+		}
+	}
+	keyFollower(mp.Value)
+
+	return rv, nil
+}
+
+// NewService returns a new Service struct derived from an *ast.TypeSpec with a
+// Type of *ast.InterfaceType.
```
hasAdamr: Any `*ast.TypeSpec` of Type `*ast.InterfaceType`?

Above it is mentioned that the `(SvcDefName)Client` interface will be passed to this.

Consider

```
NewService returns a new Service from an *ast.TypeSpec with a Type of "{SVCNAME}Client" interface 
```

svcdef/svcdef.go
```diff
@@ -0,0 +1,354 @@
...
+		}
+		rv.Methods = append(rv.Methods, nmeth)
+	}
+	return rv, nil
+}
+
+// NewServiceMethod returns a new ServiceMethod derived from the provided
+// *ast.Field. The given *ast.Field is intended to have a Type of *ast.FuncType
+// from an *ast.InterfaceType's Methods.List attribute. Providing an *ast.Field
+// with a different structure may return an error.
```
hasAdamr: Consider

```
NewServiceMethod returns a new ServiceMethod derived from a method of a Service interface. This is accepted in the form of an *ast.Field which contains the name of the method.
```

svcdef/svcdef.go
```diff
@@ -0,0 +1,354 @@
...
+	ft, ok := m.Type.(*ast.FuncType)
+	if !ok {
+		return nil, errors.New("Provided *ast.Field.Type is not of type *ast.FuncType; cannot proceed")
+	}
+
+	input := ft.Params.List
+	output := ft.Results.List
+
+	// Zero'th param of a serverMethod is Context.context, while first param is
+	// this methods RequestType
```
hasAdamr: Possibly add the function signature to show what params and returns you are choosing.

svcdef/svcdef_test.go
```diff
@@ -0,0 +1,131 @@
+package svcdef
+
+import (
+	"io"
+	"os"
+	"strings"
+	"testing"
+)
+
+func TestNewCatalog(t *testing.T) {
```
hasAdamr: TestNewSvcDef? old naming making its way though.

svcdef/consolidate_http.go
```diff
@@ -0,0 +1,132 @@
...
+			}
+		}
+		return nil
+	}
+
+	// Since this functionality must be repeated for each HTTPBinding on each
+	// ServiceMethod, and since this logic would be a bit hard to make sense of
+	// nested in a for loop, it's been broken out into this function. Since
+	// it's only used within this function, and is purely to make this logic
+	// more readable, I chose to define it within this function.
```
hasAdamr: What is "it"

> I chose to define **it** within this function

Consider

```
// This logic has been broken out of the for loop below to flatten 
// this function and avoid difficult to read nesting
```

svcdef/svcdef.go
```diff
@@ -0,0 +1,354 @@
...
+				case *ast.TypeSpec:
+					rv = append(rv, ts)
+				}
+			}
+		}
+	}
+	return rv, nil
+}
+
+// New creates a Svcdef by parsing the provided Go and Protobuf source files to
```
hasAdamr, line 136: I like this comment a lot. Nice job++

svcdef/svcdef.go
```diff
@@ -0,0 +1,354 @@
...
+	for _, gofile := range goFiles {
+		fset := token.NewFileSet()
+		fileAst, err := parser.ParseFile(fset, "", gofile, parser.ParseComments)
+		if err != nil {
+			return nil, errors.Wrap(err, "couldn't parse go file to create Svcdef")
+		}
+
+		typespecs, _ := retrieveTypeSpecs(fileAst)
+		for _, t := range typespecs {
+			//sp.Dump(t)
```
hasAdamr: Clean

svcdef/svcdef.go
```diff
@@ -0,0 +1,354 @@
...
+//
+//     1. The KeyType of the *ast.MapType will always be an ast.Ident
+//     2. The ValueType may be an ast.Ident OR an ast.StarExpr -> ast.Ident
+//
+// These rules are a result of the rules for map fields of Protobuf messages,
+// namely that a key may only be represented by a non-float basetype (e.g.
+// int64, string, etc.), and that a value may be either a basetype or a Message
+// type or an Enum type. In the resulting Go code, a basetype will be
+// represented as an ast.Ident, while a key that is a Message or Enum type will
+// be represented as an *ast.StarExpr which references an ast.Ident.
```
hasAdamr, line 228: > //This code cannot accept an arbitrary
> // MapType, only one which follows the conventions of Go code generated by
> // protoc-gen-go.

It might be a good idea to add a package level description of svcdef saying

```
// package svcdef provides types used to represent a grpc Service with protobuf messages. 
// It implements a parser for Go code generated by protoc-gen-go and all methods accept 
// only ast of the form created by protoc-gen-go.
```

Then include these specifics within the function body to be read by someone working on the function itself.

svcdef/consolidate_http.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return "", ""
+}
+
+// paramLocation returns the location that a field would be found according to
+// the rules of a given HTTPBinding.
+func paramLocation(field *Field, binding *svcparse.HTTPBinding) string {
+	path_params := getPathParams(binding)
+	for _, path_param := range path_params {
+		if strings.Split(path_param, ".")[0] == field.Name {
```
hasAdamr, line 101: Why is this split on a `.`?

hasAdamr (commented):
Looking good. Left mostly comment reviews. Overall looking good. 

Would be good to add a package level description, I mention this in my comment on NewMap.

Looking forward to more robust type handling in truss.

svcdef/resolvetypes.go
hasAdamr: So if I do 

```
        for i, f := range m.Fields {
            fmt.Println(f, " : f")
            fmt.Println(m.Fields[i], " : m.Fields[i]")
            setType(f.Type, tmap)
```

Then the addresses that print are all equal.

`m.Fields` is a `[]*Field` so in your table it is a `[n]E` where `E` is a `*Field` so `a[i]` is a `*Field`.

svcdef/consolidate_http.go
hasLeland: That behavior is an undocumented hack, as it turns out...

I should point out that most of the `consolidateHTTP` code is taken straight from `doctree`, so this code was written relatively early in the life of truss.

The point of this line of code is to paper over a not-often-used feature that I didn't know how to implement, namely specifying the location of sub-parameters of messages in the path. Here's an example:

```
message MsgA {
    int a = 1;
    MsgB b = 2;
}
message MessageB {
    int Foo = 1;
}
service Exmpl {
  rpc GetA (MsgA) returns (MsgA) {
    option (google.api.http) = {
      get: "/{b.Foo}"
    };
  }
}
```

The above is totally valid, and _eventually_ we should support this. But for right now, I've not implemented this. Instead, the gross hack that I have above would make svcdef thing that just `b` was embedded in the path.

I've added a comment noting this incompatibility.

PR #73, opened 2016-10-22
Cleanup templates

hasAdamr:
Remove code from templates that generate commented out code in generated service.

Originally this code was left in as I was imagining that we would uncomment it as we understood it better.

Turns out that it mostly causes clutter and makes working with the templates harder than it needs to be.

If we ever want to see that commented out code, we can always look at go-kit's examples as that is where it originated from.


zaquestion (approved):

zaquestion:
A++++++++

hasLeland:
An excellent change!

## Looks good to me! :+1:

PR #74, opened 2016-10-24
Pipeline migration -- Part 1

hasLeland:
This is part 1 of a proposed piecemeal and transparent migration of most of the infrastructure of Truss off of `doctree` and onto `svcdef`.
# What changed?
1. This adds a new package called `parsepkgname` which parses protobuf files and returns their package names. This is added as a fix for the previous behavior having to create a `doctree` just to properly initialize our configuration; with this package, we can now create our `Config` struct cleanly and completely before moving on to other steps in generation.
2. These changes integrate the creation of a `Svcdef` struct from our `.proto` and `.pb.go` files. However, this newly created `Svcdef` is unused at this point. My plan is to do a piecemeal transparent migration from `doctree` to `svcdef`, and this is the first part of that. The next changes I plan are to move `gengokit` from `doctree` to `svcdef`.

Let me know what you think of the code, and let me know what you think of the idea of a piecemeal, transparent, migration to `svcdef`!

### DIFF
--- a/deftree/svcparse/scanner.go
+++ b/deftree/svcparse/scanner.go
@@ -301,6 +301,9 @@ func (self *SvcScanner) FastForward() error {
 	return nil
 }
 
+// ReadUnit returns the next "group" of runes found in the input stream. If the
+// end of the stream is reached, io.EOF will be returned as error. No other
+// errors will be returned.
 func (self *SvcScanner) ReadUnit() ([]rune, error) {
 	var rv []rune
 	var err error = nil
--- a/deftree/svcparse/scanner.go
+++ b/deftree/svcparse/scanner.go
@@ -19,7 +19,7 @@ type RuneReader struct {
 
 func (self *RuneReader) ReadRune() (rune, error) {
 	var toret rune = 0
-	var err error = nil
+	var err error
 
 	if self.RunePos < self.ContentLen {
 		toret = self.Contents[self.RunePos]
@@ -306,7 +306,7 @@ func (self *SvcScanner) FastForward() error {
 // errors will be returned.
 func (self *SvcScanner) ReadUnit() ([]rune, error) {
 	var rv []rune
-	var err error = nil
+	var err error
 	if self.UnitPos < len(self.Buf) {
 		unit := self.Buf[self.UnitPos]
 
--- a/truss/main.go
+++ b/truss/main.go
@@ -18,6 +18,7 @@ import (
 	"github.com/TuneLab/go-truss/deftree"
 	"github.com/TuneLab/go-truss/gendoc"
 	"github.com/TuneLab/go-truss/gengokit"
+	"github.com/TuneLab/go-truss/svcdef"
 )
 
 var (
@@ -43,7 +44,8 @@ func main() {
 	cfg, err := parseInput()
 	exitIfError(errors.Wrap(err, "cannot parse input"))
 
-	dt, err := parseServiceDefinition(cfg.DefPaths)
+	//dt, sd, err := parseServiceDefinition(cfg)
+	dt, _, err := parseServiceDefinition(cfg)
 	exitIfError(errors.Wrap(err, "cannot parse input definition proto files"))
 
 	genFiles, err := generateCode(cfg, dt)
@@ -82,7 +84,7 @@ func parseInput() (*truss.Config, error) {
 	if err != nil {
 		return nil, errors.Wrapf(err, "Could not open package file %q", cfg.DefPaths[0])
 	}
-	svcName, err := parsepkgname.PackageNameFromFile(defFile)
+	svcName, err := parsepkgname.FromReader(defFile)
 	if err != nil {
 		return nil, errors.Wrapf(err, "cannot parse package name from file %q", cfg.DefPaths[0])
 	}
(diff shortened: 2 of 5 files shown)


truss/parsepkgname/parsepkgname.go
```diff
@@ -0,0 +1,111 @@
...
+	case r == '/' && len(unit) > 1:
+		rv = COMMENT
+	}
+	return rv
+}
+
+// PackageNameFromFile accepts an io.Reader, the contents of which should be a
+// valid proto3 file, and returns the name of the protobuf package for that
+// file.
+func PackageNameFromFile(protofile io.Reader) (string, error) {
```
hasAdamr: To keep with Go's package style, how about

```
parsepkgname.FromFile(
```

truss/parsepkgname/parsepkgname.go
```diff
@@ -0,0 +1,111 @@
...
+	scanner := svcparse.NewSvcScanner(protofile)
+	return GetPackageName(scanner)
+}
+
+// GetPackageName accepts a Scanner for a protobuf file and returns the name of
+// the protobuf package that the file lives within.
+func GetPackageName(scanner Scanner) (string, error) {
```
hasAdamr: How about:

```
parsepkgname.FromScanner(
```

truss/parsepkgname/parsepkgname.go
```diff
@@ -0,0 +1,111 @@
...
+// PackageNameFromFile accepts an io.Reader, the contents of which should be a
+// valid proto3 file, and returns the name of the protobuf package for that
+// file.
+func PackageNameFromFile(protofile io.Reader) (string, error) {
+	scanner := svcparse.NewSvcScanner(protofile)
+	return GetPackageName(scanner)
+}
+
+// GetPackageName accepts a Scanner for a protobuf file and returns the name of
+// the protobuf package that the file lives within.
```
hasAdamr: > the protobuf package that the file lives within.

Maybe

> the protobuf package that the file belongs to.

hasAdamr (commented):
Looks great. Seems like someone really has gotten into in-line functions, which is fun haha. 

Only really a few naming suggestions and a comment change.

deftree/svcparse/scanner.go
```diff
@@ -301,6 +301,9 @@ func (self *SvcScanner) FastForward() error {
 	return nil
 }
 
+// ReadUnit returns the next "group" of runes found in the input stream. If the
+// end of the stream is reached, io.EOF will be returned as error. No other
+// errors will be returned.
 func (self *SvcScanner) ReadUnit() ([]rune, error) {
 	var rv []rune
 	var err error = nil
```
hasIan: Style note: variables declared with `var` without explicit initialization are guaranteed to be initialized to their zero values (`nil` in the case of `error`). So the `= nil` is redundant here.

truss/main.go
```diff
@@ -79,82 +79,118 @@ func parseInput() (*truss.Config, error) {
...
 
-// parseServiceDefinition returns a deftree which contains all needed for all
-// generating a truss service and documentation
-func parseServiceDefinition(definitionPaths []string) (deftree.Deftree, error) {
-	protocOut, err := execprotoc.CodeGeneratorRequest(definitionPaths)
-	if err != nil {
-		return nil, errors.Wrap(err, "cannot use parse input files with protoc")
+// parseServiceDefinition returns a deftree which contains all necessary
+// information for generating a truss service and its documentation.
+func parseServiceDefinition(cfg *truss.Config) (deftree.Deftree, *svcdef.Svcdef, error) {
```
hasIan, line 120: Is the thinking here that eventually this will just return a `Svcdef`? Is the `Deftree` here just for compatibility with the current state of Truss? Am I understanding your intent correctly?

truss/main.go
```diff
@@ -79,82 +79,118 @@ func parseInput() (*truss.Config, error) {
...
-	svcFile, err := execprotoc.ServiceFile(protocOut, filepath.Dir(definitionPaths[0]))
+	err := execprotoc.GeneratePBDotGo(cfg.DefPaths, svcPath, cfg.PBPath())
 	if err != nil {
-		return nil, errors.Wrap(err, "cannot find service definition file")
+		return nil, nil, errors.Wrap(err, "cannot create .pb.go files")
 	}
 
-	dt, err := deftree.New(protocOut, svcFile)
+	// Open all .pb.go files and store in slice to be passed to svcdef.New()
+	var openFiles func([]string) ([]io.Reader, error)
```
hasIan: You don't need this line if you just write the next line with the short declaration operator.

```
openFiles := func(paths []string) ([]io.Reader, error) {
// ...
}
```

truss/main.go
```diff
@@ -79,82 +79,118 @@ func parseInput() (*truss.Config, error) {
...
+	openFiles = func(paths []string) ([]io.Reader, error) {
```
hasIan: This function isn't using any closures--it's going to be the same every time the enclosing function runs. In cases like these I prefer declaring functions in "traditional" fashion. I have found, though, that this is a stylistic preference that not everyone shares. I suppose I like the clarity of knowing right away that a function does not close over values in the enclosing function, which is made explicit by being declared outside the calling function's body.

truss/parsepkgname/parsepkgname.go
hasIan: Or it could work with a less specific package name. As it is, though, it stutters.

hasIan: Additionally, it is called `PackageNameFromFile`, but it takes an `io.Reader`. The function name is overly specific.

truss/parsepkgname/parsepkgname.go
```diff
@@ -0,0 +1,111 @@
...
+	"io"
+	"unicode"
+
+	"github.com/TuneLab/go-truss/deftree/svcparse"
+)
+
+type Token int
+
+const (
+	IDENT Token = iota
```
hasIan: Is it important that these constants are exported? Idiomatic Go shies away from `ALL_CAPS_IDENTIFIERS`. I would have named them something like `tokenIdent`, `tokenWhitespace`, `tokenComment`, `tokenOther`. Prefixes are pretty Go-y (witness all the `ErrSomeError` errors stuff).

If the tokens _should_ be exported, then never mind this; I'm not going to freak out about all-caps identifiers. Not too much, anyway.

truss/parsepkgname/parsepkgname.go
hasAdamr: FromReader then?

truss/parsepkgname/parsepkgname.go
```diff
@@ -0,0 +1,111 @@
...
+	// If you need a scanner which provides these out of the box, see the
+	// SvcScanner struct in github.com/TuneLab/go-truss/deftree/svcparse
+	ReadUnit() ([]rune, error)
+}
+
+func categorize(unit []rune) Token {
+	rv := OTHER
+	r := unit[0]
+	switch {
+	case unicode.IsLetter(r):
```
hasIan: Go supports multiple switch cases on each case line. You can collapse this switch down to the following:

``` go
switch {
case unicode.IsLetter(r), unicode.IsDigit(r), r == '_':
    return IDENT
case unicode.IsSpace(r):
    return WHITESPACE
case r == '/' && len(unit) > 1:
   return COMMENT
}
return OTHER
```

truss/main.go
hasAdamr: Deftree will still be needed for generating documentation down the line. Currently though, Svcdef is not being used for anything. This is just the pull getting the creation of an Svcdef into the pipeline.

truss/parsepkgname/parsepkgname.go
```diff
@@ -0,0 +1,111 @@
...
+		rv = IDENT
+	case unicode.IsDigit(r):
+		rv = IDENT
+	case r == '_':
+		rv = IDENT
+	case unicode.IsSpace(r):
+		rv = WHITESPACE
+	case r == '/' && len(unit) > 1:
```
hasIan: I'm not entirely sure where the `unit` comes from, but just a sanity check; this is a legitimate test for a comment?

truss/parsepkgname/parsepkgname.go
```diff
@@ -0,0 +1,111 @@
...
+// GetPackageName accepts a Scanner for a protobuf file and returns the name of
+// the protobuf package that the file lives within.
+func GetPackageName(scanner Scanner) (string, error) {
+	foundpackage := false
+
+	// A nice way to ignore comments. Recursively calls itself until it
+	// recieves a unit from the scanner which is not a comment.
+	var readIgnoreComment func(Scanner) (Token, []rune, error)
+	readIgnoreComment = func(scn Scanner) (Token, []rune, error) {
+		unit, err := scanner.ReadUnit()
```
hasIan: Here, you _are_ closing over the `scanner` value, _but_ you are also declaring a parameter `scn`, and using both. Are you intentionally using two scanners here?

truss/parsepkgname/parsepkgname.go
```diff
@@ -0,0 +1,111 @@
...
+		tkn := categorize(unit)
+		if tkn == COMMENT {
+			return readIgnoreComment(scn)
+		}
+		return tkn, unit, err
+	}
+
+	for {
+		tkn, unit, err := readIgnoreComment(scanner)
+		// Err may only be io.EOF
```
hasIan: Instead of this comment, why not explicitly return `io.EOF`, if we guarantee that no other error can ever be returned from this function (which is how I'm reading this comment)? Of course, that leaves the actual `err` value untouched here. Will _it_ always be `io.EOF`? Something about this construct is a little weird. Not as a big a deal as it sounds like I'm making it here, but a little weird.

truss/parsepkgname/parsepkgname_test.go
```diff
@@ -0,0 +1,89 @@
...
+	contents [][]rune
+	position int
+}
+
+func (t *testScanner) ReadUnit() ([]rune, error) {
+	if t.position < len(t.contents) {
+		rv := t.contents[t.position]
+		t.position += 1
+		return rv, nil
+	} else {
```
hasIan: You don't need an `else` here. `golint` will catch stylistic stuff like this for you.

```
➜  dummy grep -A 10 "func even" main.go
func even(i int) bool {
    if i%2 == 0 {
        return true
    } else {
        return false
    }
}
➜  dummy golint main.go
main.go:15:12: if block ends with a return statement, so drop this else and outdent its block
```

(I know explicitly returning `true` or `false` here is dum dum; it was just the first example I thought of.)

hasIan:
Overall, looks really good. I added some comments stream-of-consciousness. Some are more salient than others.

truss/parsepkgname/parsepkgname.go
hasLeland: I will be going with your `parsepkgname.FromScanner(`, seems like the clearest name.

truss/parsepkgname/parsepkgname.go
hasLeland: That is definitely an error, thank you for pointing that out!

truss/parsepkgname/parsepkgname.go
hasLeland: What I was attempting to convey at the time was that if a non-nil error where to be returned from `readIgnoreComment()`, then that error will only ever be `io.EOF`. However, that information isn't particularly relevant to anyone, so I'm going to remove the comment.

hasLeland:
I'm pretty sureI've addressed all the comments shared here. As a result, I wil merge this branch at 10am.

PR #75, opened 2016-10-26
Context to transport to transport to context

hasAdamr:
This pull request adds:
- **Server**: Put all headers from incoming HTTP requests into the context
- **ClientLib**: Configuration to take values from the context and put them into outgoing HTTP request headers
- Tests for these two cases

Unfortunately, a lot of this code is in template form, which may be hard to review. 

[Server Code](https://github.com/TuneLab/go-truss/compare/master...hasAdamr:context-to-transport-to-transport-to-context?expand=1#diff-0fe7c6aec1f529107babf5f63bddb542R322)

[Client Code](https://github.com/TuneLab/go-truss/compare/master...hasAdamr:context-to-transport-to-transport-to-context?expand=1#diff-0fe7c6aec1f529107babf5f63bddb542R358)


gengokit/httptransport/httptransport.go
```diff
@@ -24,6 +24,8 @@ import (
 type Helper struct {
 	Methods           []*Method
 	PathParamsBuilder string
+	ServerTemplate    func(interface{}) (string, error)
+	ClientTemplate    func(interface{}) (string, error)
```
hasLeland, line 28: This isn't really a critique; I'm curious why `ServerTemplate` and `ClientTemplate` aren't methods of `Helper`? I'm assuming this was done because neither of them actually depends on `Helper`, so this keeps them isolated?

gengokit/httptransport/templates.go
```diff
@@ -3,102 +3,102 @@ package httptransport
...
-		return nil, errors.Wrapf(err, "Error while reading query params: %v", r.URL.Query())
+{{- with $binding := . -}}
+	// DecodeHTTP{{$binding.Label}}Request is a transport/http.DecodeRequestFunc that
+	// decodes a JSON-encoded {{ToLower $binding.Parent.Name}} request from the HTTP request
+	// body. Primarily useful in a server.
+	func DecodeHTTP{{$binding.Label}}Request(_ context.Context, r *http.Request) (interface{}, error) {
+		var req pb.{{GoName $binding.Parent.RequestType}}
+		err := json.NewDecoder(r.Body).Decode(&req)
+		// err = io.EOF if r.Body was empty
+		if err != nil && err != io.EOF {
```
hasLeland, line 14: Why have these templates all been indented? Aren't they going to be unindented in the actual code?

hasLeland:
Thinking about it, the current way of rendering the templates for client and server, while a clever implementation, seems a bit fragile.

The recursive template execution is more than a bit confusing, and I can easily seeing it biting us in the future. To avoid this structural recursion, I think that `httptransport` should be rearranged to better reflect what is now actually going on.

I think http transport should pretty much only expose a `RenderClient` and a `RenderServer` method on a `Helper` struct. Then, when the `Render*` methods are called, the existing code for figuring out all the http-transport code to generate is called transparently beneath.

## Diagrams

Here's what the recursive structure looks like now:

```
+────────────────────────────────────────+
│                                        │
+──> .                                   │
     └── httptransport                   │
         ├── client─server─creation      │
         │   └───────────────────────────+
         └── transport+generation
```

Here's what I propose it should probably look like:

```
    .
    └── httptransport
        └── client-server-creation
            └── transport-generation
```

Whether this means having sub-packages of `httptransport` or just a code restructure, I don't know. Additionally, I don't think this is urgent, but I want to at least write this down somewhere so I don't forget : )

hasLeland:
In general though, I really like these changes. They're not huge, and they're not incomprehensible. I especially like the straightforwardness of the tests for this ctx-to-ctx behavior.

Looks really good @hasAdamr, good job!

gengokit/httptransport/httptransport.go
hasAdamr: They do not operate on the struct or use any information from the struct so they are func fields of the struct rather than methods of the struct.

gengokit/httptransport/templates.go
hasAdamr: They have been indented for readablity.

gofmt is run on everything before it is added to the returned files slice in gengokit, so horozontal whitespace in the templates will always be flattened.

PR #76, opened 2016-10-28
Add example and update README.md

hasAdamr:
Adding an example .proto file for someone who wants to try truss see what is up.

Also updated our install instructions to be clearer and more actionable.


hasLeland:
## Looks goot do me!

PR #77, opened 2016-10-29
gRPC context

hasAdamr:
Add context to transport to transport to context for gRPC transport using metadata.
Update integration tests to test grpc transport, specifically this addition.

package http calls http.CanonicalHeaderKey on each key over transport and grpc metadata calls strings.ToLower on each key over transport. My solution was to add both of these forms of the key to the context, so that the value can be accessed in either form.

i.e.

```
ctx = context.WithValue(context.Background(), "Truss-test", "SECRET")
```

On the server side, `"SECRET"` is accessible via either

```
ctx.Value("Truss-Test")
// or
ctx.Value("truss-test")
// BUT NOT
ctx.Value("Truss-test")
```

Comments about this compromise is appreciated.


gengokit/httptransport/templates.go
```diff
@@ -321,7 +321,12 @@ func EncodeHTTPGenericResponse(_ context.Context, w http.ResponseWriter, respons
 
 func headersToContext(ctx context.Context, r *http.Request) context.Context {
 	for k, _ := range r.Header {
+		// The key is added both in http format (k) which has had
+		// http.CanonicalHeaderKey called on it in transport as well as the
+		// strings.ToLower which is the grpc metadata format of the key so
+		// that it can be accessed in either format
```
hasLeland, line 327: Consistency A++ :+1:

gengokit/template/NAME-service/generated/client/grpc/client.gotemplate
```diff
@@ -14,11 +17,21 @@ import (
 )
 
 // New returns an service backed by a gRPC client connection. It is the
-// responsibility of the caller to dial, and later close, the connection.
-func New(conn *grpc.ClientConn) handler.Service {
-	//options := []grpctransport.ClientOptions{
-		//grpctransport.ClientBefore(),
-	//}
+// responsibility of the caller to dial, and later close, the connection.  
+func New(conn *grpc.ClientConn, options ...ClientOption) (handler.Service, error) {
```
hasLeland, line 21: Yay for now passing errors!

hasLeland:
## Looks good to me!

PR #80, opened 2016-10-31
Migrate gengokit to svcdef

hasLeland:
# What?

1. This is a complete migration of gengokit to `svcdef` and away from `deftree`.
2. Fixes a bug causing `svcdef` to return an error when parsing a proto file without http annotations. Instead, will print a warning before proceding.
3. Adds a `FromString()` function to `svcdef`, very useful for testing!
4. Adds and cleans up a bunch of tests in a bunch of places

# What does this *not* cover?

This pull request migrates gengokit to `svcdef`, but it does not refactor packages like `httptransport` and `clientarggen` to leverage the benefits of `svcdef`. However, once this is merged I will begin work to improve those packages by using the improvements in `svcdef`.


svcdef/consolidate_http.go
```diff
@@ -22,13 +25,13 @@ func consolidateHTTP(sd *Svcdef, protoFiles []io.Reader) error {
 		lex := svcparse.NewSvcLexer(pfile)
 		protosvc, err := svcparse.ParseService(lex)
 		if err != nil {
-			if strings.Contains("'options' or", err.Error()) {
+			if strings.Contains(err.Error(), "'option' or 'additional_bindings'") {
```
hasAdamr: Because we are using `pkg/errors` I believe we want an `err.Cause().Error()` here. 

Also how does the "'this' or 'that'" syntax work for `strings.Contains`?

gengokit/gengokit.go
```diff
@@ -102,24 +99,6 @@ func GenerateGokit(dt deftree.Deftree, goPackage, goPBPackage string, previousFi
 	return codeGenFiles, nil
 }
 
-// getProtoService finds returns the service within a deftree.Deftree
```
hasAdamr, line 105: Awesome job getting rid of this concern from gengokit.

gengokit/config/config.go
```diff
@@ -0,0 +1,10 @@
+package config
```
hasAdamr, line 1: :thumbsup:

gengokit/httptransport/httptransport.go
```diff
@@ -12,15 +12,14 @@ import (
...
-	"github.com/TuneLab/go-truss/gengokit/clientarggen"
+	"github.com/TuneLab/go-truss/svcdef"
 	gogen "github.com/golang/protobuf/protoc-gen-go/generator"
 	"github.com/pkg/errors"
 )
 
 // Helper is the base struct for the data structure containing all the
 // information necessary to correctly template the HTTP transport functionality
-// of a microservice. Helper must be built from a deftree.
+// of a microservice. Helper must be built from a Svcdef.
```
hasAdamr: I think we should follow what Ian said and try to switch to the word "service" rather than "microservice". It has less baggage from being a buzz word. I have been changing them when I run across them. 

If you make more modifications to this pull, consider changing it here.

gengokit/httptransport/types.go
```diff
@@ -49,12 +50,14 @@ type Field struct {
...
 	Location string
 	// The protobuf type that this field is of.
+	// Deprecated by Svcdef.
 	ProtobufType string
 	// The type within the Go language that's used to represent the original
 	// field that this field refers to.
 	GoType string
 	// The protobuf label for the original field that this field refers to. Is
 	// probably "OPTIONAL", though may be "REPEATED".
+	// Deprecated by Svcdef.
```
hasAdamr: Consider all caps "DEPRECATED" so it is very visable.

svcdef/consolidate_http.go
```diff
@@ -55,12 +58,13 @@ func assembleHTTPParams(svc *Service, httpsvc *svcparse.Service) error {
...
 
-		params := make([]*HTTPParameter, 0)
+		//params := make([]*HTTPParameter, 0)
+		var params []*HTTPParameter
 		for _, field := range msg.Fields {
-			new_param := &HTTPParameter{}
-			new_param.Field = field
-			new_param.Location = paramLocation(field, parsedbind)
-			params = append(params, new_param)
+			newParam := &HTTPParameter{}
```
hasAdamr, line 73: Look at the change to golang variable name style. :+1:

svcdef/newfromstring.go
```diff
@@ -0,0 +1,59 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/TuneLab/go-truss/truss/execprotoc"
+	"github.com/pkg/errors"
+)
+
+// NewFromString creates a Svcdef from a string of a valid protobuf file. Very
+// useful in tests.
+func NewFromString(def string) (*Svcdef, error) {
```
hasAdamr, line 16: :+1:

svcdef/newfromstring_test.go
```diff
@@ -0,0 +1,173 @@
...
+		},
+	}
+	output := sd.Service.Methods[0].Bindings
+	if got, want := output, expected; !reflect.DeepEqual(got, want) {
+		diff := gentesthelper.DiffStrings(spew.Sdump(got), spew.Sdump(want))
+		t.Errorf("got != want; methods differ: %v\n", diff)
+	}
+}
+
+func TestNoHTTPBinding(t *testing.T) {
```
hasAdamr, line 146: :+1:

svcdef/svcdef.go
```diff
@@ -144,8 +146,12 @@ func New(goFiles []io.Reader, protoFiles []io.Reader) (*Svcdef, error) {
 		if err != nil {
 			return nil, errors.Wrap(err, "couldn't parse go file to create Svcdef")
 		}
+		rv.PkgName = fileAst.Name.Name
 
-		typespecs, _ := retrieveTypeSpecs(fileAst)
+		typespecs, err := retrieveTypeSpecs(fileAst)
```
hasAdamr, line 151: Glad this error is handled now :).

svcdef/svcdef.go
```diff
@@ -373,3 +383,14 @@ func NewField(f *ast.Field) (*Field, error) {
 	}
 	return rv, nil
 }
+
+// isExported returns true if the provided name of a declaration begins with a
+// capital letter.
+func isExported(name string) bool {
+	// Abuse DecodeRuneInString to get the first rune in the string
```
hasAdamr: Abuse is a pretty colorful word. I don't think it belongs. 

Also how is this an abuse? The standard library does something similar. https://github.com/golang/go/blob/832082b44e2965c136e53a9b8009d2e860766000/src/cmd/compile/internal/gc/export.go#L51

hasAdamr (approved):

gengokit/httptransport/types.go
hasLeland: Ha, at the time I wrote that for my own benefit with the intention of erasing those field members entirely once I'd moved on. So I'm going to just remove those, as they're not used anywhere.

svcdef/svcdef.go
hasLeland: You're right, removing that comment.

svcdef/consolidate_http.go
hasLeland: I don't know what you mean when you say "'this' or 'that'"? There's not any special syntax at play, the error I'm trying to handle in this case has a string value of:

```
parser expected 'option' or 'additional_bindings' while parsing options in line '20', instead found '}'
```

Also, I'm implementing the `errors.Cause()` behavior because that would be awesome!

svcdef/consolidate_http.go
```diff
@@ -25,7 +34,9 @@ func consolidateHTTP(sd *Svcdef, protoFiles []io.Reader) error {
 		lex := svcparse.NewSvcLexer(pfile)
 		protosvc, err := svcparse.ParseService(lex)
 		if err != nil {
-			if strings.Contains(err.Error(), "'option' or 'additional_bindings'") {
+			fmt.Println(err.Error())
```
hasAdamr: Do we want this to both print the error in this way and print out the warning? I'm fine with that, just want to make sure this is not for debugging.

hasLeland: Ah whoops, that was something I left in from debuging.

Issue #72, opened 2016-10-15
Multiple multiline comments in a service definition causes http transport to fail


adamryman:
Reproduce:

Call truss on this file

```
syntax = "proto3";

package foo;

import "google/api/annotations.proto";

service Foo {
  /*rpc NoFoo(NoFoo) returns (NoFoo) {*/
    /*option (google.api.http) = {*/
      /*get: "/foo/{FooId}"*/
    /*};*/
  /*}*/
  rpc FooBarer(FooBar) returns (FooBar) {
    option (google.api.http) = {
      get: "/foo"
    };
  }
}

message FooBar {
  int64 FooId = 1;
}
```

```
WARN[0000] Error found while parsing file foo.proto      
error=parser expected identifier 'rpc' in line '9', instead found '/**/'
```

hasLeland:
Fixed in #81

PR #81, opened 2016-10-31
Fix bug #72, add tests

hasLeland:
This adds a fix to the `deftree/svcdef` parser for #72.

### DIFF
--- a/deftree/svcparse/lexer.go
+++ b/deftree/svcparse/lexer.go
@@ -68,7 +68,9 @@ func NewTokenGroup(scn *SvcScanner) *TokenGroup {
 					}
 					twostr := string(two)
 					if len(two) > 1 && two[0] == '/' {
-						str += twostr
+						// Ensure that whitespace between multiline comments on
+						// a single line are preserved.
+						str += onestr + twostr
 					} else {
 						scn.UnReadToPosition(one_pos)
 						break
--- a/deftree/svcparse/lexer_test.go
+++ b/deftree/svcparse/lexer_test.go
@@ -155,6 +155,41 @@ func TestLexSingleLineComments(t *testing.T) {
 			t.Fatalf("%v returned token '%v' differs from expected token '%v'\n", i, cleanStr(str), cleanStr(good_str))
 		}
 
+		if tk == EOF {
+			break
+		}
+	}
+}
+
+func TestLexMultiLineComments(t *testing.T) {
+	r := strings.NewReader("service testing\n /* comment1 */ /* thing */\n/*comment2 */\n\n/*comment 3 */\n what")
+	lex := NewSvcLexer(r)
+	for i, good_str := range []string{
+		"service",
+		" ",
+		"testing",
+		"\n ",
+		"/* comment1 */ /* thing */",
+		"\n",
+		"/*comment2 */",
+		"\n\n",
+		"/*comment 3 */",
+		"\n ",
+		"what",
+		"",
+	} {
+		tk, str := lex.GetToken()
+		if tk == ILLEGAL {
+			t.Fatalf("Recieved ILLEGAL token on '%v' call to GetToken\n", i)
+		}
+
+		if str != good_str {
+			for _, grp := range lex.Buf {
+				t.Logf("  '%v'\n", cleanStr(grp.value))
+			}
+			t.Fatalf("%v returned token '%v' differs from expected token '%v'\n", i, cleanStr(str), cleanStr(good_str))
+		}
+
 		if tk == EOF || tk == ILLEGAL {
 			break
 		}
--- a/deftree/svcparse/parser_test.go
+++ b/deftree/svcparse/parser_test.go
@@ -527,3 +527,49 @@ service FlowCombination {
 		t.Errorf("Http binding contents = %#v, want = %#v\n", got, want)
 	}
 }
+
+func TestOddComments(t *testing.T) {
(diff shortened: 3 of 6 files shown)


deftree/svcparse/lexer.go
```diff
@@ -68,7 +68,7 @@ func NewTokenGroup(scn *SvcScanner) *TokenGroup {
 					}
 					twostr := string(two)
 					if len(two) > 1 && two[0] == '/' {
-						str += twostr
+						str += onestr + twostr
```
hasAdamr: I don't understand the surrounding material enough to know, but this feels like it deserves a comment.

deftree/svcparse/lexer_test.go
```diff
@@ -161,6 +161,41 @@ func TestLexSingleLineComments(t *testing.T) {
 	}
 }
 
+func TestLexMultiLineComments(t *testing.T) {
```
hasAdamr: :+1: :+1: :+1:

deftree/svcparse/parser_test.go
```diff
@@ -527,3 +527,49 @@ service FlowCombination {
 		t.Errorf("Http binding contents = %#v, want = %#v\n", got, want)
 	}
 }
+
+func TestOddComments(t *testing.T) {
```
hasAdamr, line 531: :+1:

deftree/svcparse/scanner.go
```diff
@@ -153,6 +153,8 @@ func BuildScanUnit(rr *RuneReader) (*ScanUnit, error) {
 						buf = append(buf, ch)
 						return setReturn(), nil
 					}
+				} else {
```
hasAdamr, line 156: Once again, maybe deserves a comment, as this fixes something but I have no idea why.

deftree/svcparse/scanner_test.go
```diff
@@ -49,3 +49,39 @@ func TestScanSingleLineComments(t *testing.T) {
 
 	}
 }
+
+func TestScanMultiLineComments(t *testing.T) {
```
hasAdamr, line 53: :+1: :+1: :+1:

deftree/svcparse/scanner_test.go
```diff
@@ -49,3 +49,39 @@ func TestScanSingleLineComments(t *testing.T) {
...
+			for _, grp := range scn.Buf {
+				t.Logf("  '%v'\n", cleanStr(string(grp.Value)))
+			}
+			t.Fatalf("%v returned token '%v' differs from expected token '%v'\n", i, cleanStr(str), cleanStr(good_str))
+		}
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			t.Fatalf(fmt.Sprintf("%v", err))
```
hasAdamr: Why a `fmt.Sprintf` inside a `t.Fatalf`? You could just put the error and the `"%v"`.

Also, don't we want a `"\n"` here?

deftree/svcparse/parser_test.go
```diff
@@ -527,3 +527,49 @@ service FlowCombination {
...
+	}
+	/* lots */ /* of */ /* comments */
+}
+	`)
+
+	lex := NewSvcLexer(r)
+	svc, err := ParseService(lex)
+
+	if err != nil {
+		t.Error(err)
```
hasAdamr, line 570: This will not cause the test to fail, just print an error. Not sure if that is what you wanted.

deftree/svcparse/lexer_test.go
```diff
@@ -161,6 +161,41 @@ func TestLexSingleLineComments(t *testing.T) {
...
+
+		if str != good_str {
+			for _, grp := range lex.Buf {
+				t.Logf("  '%v'\n", cleanStr(grp.value))
+			}
+			t.Fatalf("%v returned token '%v' differs from expected token '%v'\n", i, cleanStr(str), cleanStr(good_str))
+		}
+
+		if tk == EOF || tk == ILLEGAL {
+			break
```
hasAdamr: Should this error or fail?

hasAdamr (commented):
Looking good. Excited to see this bug be fixed. Make sure to close the issue once you address this feedback. :)

deftree/svcparse/lexer_test.go
hasLeland: What should actually happen there is that there shouldn't be an `or`, it should just be:

```
if tk == EOF {
    break
}
```

deftree/svcparse/parser_test.go
hasLeland: From the docs:

```
func (*T) Errorf
  func (c *T) Errorf(format string, args ...interface{})
    Errorf is equivalent to Logf followed by Fail.
```

`t.Error()` will cause the test to fail, but it won't halt the test.

deftree/svcparse/scanner_test.go
hasLeland: You ask this as though I didn't just copy-paste most of this test from another test, one which was written months ago at the start of my Go knowledge : )

But yes, I should change those.

deftree/svcparse/lexer.go
hasLeland: What do you mean by "deserves a comment"? As in I should comment the code, or I should explain it here?

hasAdamr: In the code? I don't have any idea what is going on so I can't say either way.

PR #82, opened 2016-11-03
Add -svcout flag to specify generated service directory

hasAdamr:
I have added a flag to specify where the `NAME-service/` directory is created, rather than always being directly to the protofiles that define the service.

### DIFF
--- a/truss/main.go
+++ b/truss/main.go
@@ -23,7 +23,8 @@ import (
 )
 
 var (
-	pbPackageFlag = flag.String("pbout", "", "The go package path where the protoc-gen-go .pb.go structs will be written.")
+	pbPackageFlag  = flag.String("pbout", "", "The go package path where the protoc-gen-go .pb.go structs will be written.")
+	svcPackageFlag = flag.String("svcout", "", "The go package path where the generated service directory will be written.")
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


truss/main.go
```diff
@@ -106,9 +117,9 @@ func parseInput() (*truss.Config, error) {
 		cfg.PBPackage = cfg.ServicePackage
 	} else {
 		cfg.PBPackage = *pbPackageFlag
-		if !fileExists(
-			filepath.Join(cfg.GOPATH, "src", cfg.PBPackage)) {
-			return nil, errors.Errorf(".pb.go output package directory does not exist: %q", cfg.PBPackage)
+		PBPackagePath := filepath.Join(cfg.GOPATH, "src", cfg.PBPackage)
+		if !fileExists(PBPackagePath) {
+			return nil, errors.Errorf(".pb.go output package directory does not exist: %q", PBPackagePath)
```
hasLeland, line 122: I like the breaking this variable out so we can get error messages with the absolute path.

truss/main.go
```diff
@@ -156,17 +167,17 @@ func parseServiceDefinition(cfg *truss.Config) (deftree.Deftree, *svcdef.Svcdef,
...
 	if err != nil {
-		return nil, nil, errors.Wrap(err, "Failed to open a .proto file")
+		return nil, nil, errors.Wrap(err, "cannot open all .proto files")
 	}
 
 	// Create the svcdef
 	sd, err := svcdef.New(pbgoFiles, pbFiles)
 	if err != nil {
-		return nil, nil, errors.Wrap(err, "Failed to create svcdef")
+		return nil, nil, errors.Wrap(err, "cannot create svcdef")
```
hasLeland, line 180: These edits to the error messages are a nice refinement.

hasLeland:
Seems like a straightforward and welcome addition.

## Looks good to me!

PR #83, opened 2016-11-04
Fix transport tests after removing handler service interface

hasAdamr:
These tests copy a handler file that had old generated code in it.

The file was not updated after everything was switched to the .pb.go
interface.


hasLeland:
Glad to see this fix for the transport tests. Once you've gotten around the problems with the gengokit/grpc update causing  the CI to fail, I think it's ready for merging.

PR #85, opened 2016-11-04
Get client arguments buildable with repeated types

hasLeland:
Prior to this, `gengokit/clientarggen` would potentially generate invalid code in the client for parsing flags to set repeated fields. This PR fixes those bugs (for basic repeated types at least) and enables the integration tests for end-to-end cli repeated fields.


gengokit/clientarggen/client_argument_generator.go
```diff
@@ -11,9 +11,10 @@ import (
 	"strings"
 	"unicode"
 
-	"github.com/TuneLab/go-truss/svcdef"
 	gogen "github.com/golang/protobuf/protoc-gen-go/generator"
 	"github.com/pkg/errors"
+
+	"github.com/TuneLab/go-truss/svcdef"
```
hasAdamr, line 17: The little things. :+1:

gengokit/httptransport/httptransport.go
```diff
@@ -95,6 +95,10 @@ func NewBinding(i int, meth *svcdef.ServiceMethod) *Binding {
 			newField.IsBaseType = true
 		}
 
+		if !newField.IsBaseType {
```
hasAdamr: Hmm, I think I prefer bringing this up to line 96 as an `else` block. 

It then convays the information. I am checking if this is not a struct type, if it is not a struct type then it is a base type, otherwise we add `pb.` to the beginning of `newField.GoType.`

gengokit/httptransport/httptransport.go
```diff
@@ -285,6 +289,9 @@ if err != nil {
 // createDecodeTypeConversion creates a go string that converts a 64 bit type to a 32 bit type
 // as strconv.ParseInt, ParseUInt, and ParseFloat always return the 64 bit type
 func createDecodeTypeConversion(f Field) string {
+	if f.Repeated {
```
hasAdamr, line 292: Possibly a very short comment about why this early exit. I can kind of understand why, but above it talks about string convert for 64 bit types and I am confused how a repeated type got in here.

hasAdamr (approved):
Two small comments, generally looks awesome. Glad this was not a lot to add.

hasAdamr:
Oh wait! One other thing. You should uncomment `// bytes N = 14;` from `cli/test-service-definitions/1-basic/`

hasLeland:
Closing due to staleness.

PR #86, opened 2016-11-08
Fix multiple protofiles causing http options parsing to catastrophically fail

hasAdamr:
Pipeline migration introduced this bug, and the test harness did not include a test for multiple protofiles so it was not caught until recently.

@hasLeland added a hotfix branch for this problem. And I have added a simple test on top of it.


hasLeland:
Glad to see these new tests, and the hot fix making its way in.

## looks good to me

PR #87, opened 2016-11-09
Move cli-client handlers to generated directory

hasAdamr:
With the updates to truss's pipeline, after pull #85 is merged in, truss now supports fully generating the cli client handlers with no user marshalling required.

This means that the `handlers/client/client_handler` file can be moved into the `generated` directory as it one more thing the user of truss does not have to worry about. It has been moved to the `generated/cli/handlers/` directory. 

`NAME-client` has also been renamed to `NAME-cli-client` to reduce confusion about the distinction between the generated cli-client and the generated client libraries.


USAGE.md
```diff
@@ -10,36 +10,34 @@ directory. The current directory should look like this:
...
+│   │   └── docs.md
+│   ├── generated
+│   │   └── ...
+│   ├── handlers
+│   │   └── server
+│   │       └── server_handler.go
+│   ├── NAME-cli-client
+│   │   └── client_main.go
+│   ├── NAME-server
+│   │   └── server_main.go
```
tunezaq, line 32: not relavent here, but still wonder if these can just be `main.go`

tunezaq (approved):

tunezaq:
Great work!

USAGE.md
hasAdamr: I see no reason why not. Do you see any reason why it is this way in the first place?

PR #92, opened 2016-11-15
Fix issue #89; cannot use gen .pb.go from multiple services

hasAdamr:
Switch from outputting `third_party` imports to using
`google.golang.org/genproto` for `http.proto` and `annotations.proto`.

This means that all generated services will import the same `http.pb.go`
and `annotations.pb.go`, fixing the problem of the proto library not
allowing for extensions to be imported from different locations.

**UPDATE:**

Based on pull request comments, this pull has been updated to search all GOPATH's for `google.golang.org/genproto` in the `protoc` invocation. It has also been updated to use the `"go/build"` package for finding the directories of the ServicePackage and the PBPackage.


deftree/build_deftree.go
```diff
@@ -92,30 +92,30 @@ func New(req *plugin.CodeGeneratorRequest, serviceFile io.Reader) (Deftree, erro
 	return &dt, nil
 }
 
-func NewFromString(def string) (Deftree, error) {
+func NewFromString(def, GOPATH string) (Deftree, error) {
```
hasIan: I know `GOPATH` is special, but in this case it's just a parameter. It looks kind of weird to me to see a regular function parameter in all caps here.

deftree/build_deftree_test.go
```diff
@@ -1,27 +1,34 @@
...
-	//"github.com/davecgh/go-spew/spew"
 
 	"github.com/golang/protobuf/proto"
 	descriptor "github.com/golang/protobuf/protoc-gen-go/descriptor"
 )
 
+var GOPATH string
+
+func init() {
+	GOPATH = filepath.SplitList(os.Getenv("GOPATH"))[0]
```
hasIan: It would be polite to inform the user whether we're ignoring any of a multi-entry GOPATH. Like, hey bub, you've got four directories in your GOPATH and we're picking the first one and ignoring the rest. If someone is using a multi-entry GOPATH and this just silently truncates it, it could be a source of angst.

README.md
```diff
@@ -1,38 +1,60 @@
...
-2. Of course, install the Go compiler and tools from https://golang.org/. See https://golang.org/doc/install for details.
-3. Install the `protoc-gen-go` and `proto` packages for Go. The simplest way is to run 
+1. Install protocol buffers version 3 or greater. The easiest way is to
+download a release from the [github releases](https://github.com/google/protobuf/releases)
+and add the binary to your `$PATH`. 
+Otherwise [install from source.](https://github.com/google/protobuf/releases)
+2. Of course, install the Go compiler and tools from https://golang.org/.
+See https://golang.org/doc/install for details.
+3. Install the `protoc-gen-go` and `proto` packages for Go. The simplest 
+way is to run
```
hasIan, line 27: Looks like we need to add `go get google.golang.org/genproto/...` into this list of things to do before Truss can be used, yeah? Or, rather, before the output of Truss can be used.

hasIan (commented):
Looks good! Minor comment about `GOPATH` looking kind of weird as a variable name, since it's not a constant, and Go doesn't even encourage all caps for constants. The `runtime` package is the only place I've really seen the all-caps thing before. Not a big deal, but it flaunts convention.

One comment about the setup documentation too, but that's it. Pretty much good to go, so far as I can tell.

deftree/build_deftree.go
adamryman: In the go standard library, they treat the "$GOPATH" this way.
```
/usr/local/go/src/go/build/build.go
...
	GOOS        string // target operating system
	GOROOT      string // Go root
	GOPATH      string // Go path
	CgoEnabled bool // whether cgo can be used
```

https://github.com/golang/go/blob/master/src/go/build/build.go#L34

If it really is a big deal. I could go `goPath`, but go does treat the "$GOPATH" standard library like this and I like going with it.

hasAdamr: In the go standard library, they treat the "$GOPATH" this way.

```
/usr/local/go/src/go/build/build.go
...
    GOOS        string // target operating system
    GOROOT      string // Go root
    GOPATH      string // Go path
    CgoEnabled bool // whether cgo can be used
```

https://github.com/golang/go/blob/master/src/go/build/build.go#L34

Though I just found this: https://github.com/golang/go/blob/master/src/cmd/go/build.go#L1790

```
for _, root := range gopath {
```

I could go with `gopath` but if it is attached to a struct I would like to use `GOPATH`

README.md
hasAdamr: That is right below this. 

``````
+   ```
+   go get -u google.golang.org/genproto    
+   ```
``````

File: deftree/build_deftree_test.go (hunk 2 above)
Comment (hasAdamr) [reply deftree/build_deftree_test.go]: I just found this: https://github.com/golang/go/blob/master/src/cmd/go/build.go#L1790

```
for _, root := range gopath {
```

I could just range over the `gopath` list. This can be used throughout truss I believe. I can start rolling in this larger change in the morning.

File: deftree/build_deftree.go (hunk 1 above)
Comment (hasIan) [reply deftree/build_deftree.go]: I don't feel too strongly either way; the mixed examples from the Go code itself satisfy me.

File: deftree/build_deftree_test.go (hunk 2 above)
Comment (hasIan) [reply deftree/build_deftree_test.go]: Sounds good.

File: README.md (hunk 3 above)
Comment (hasIan) [reply README.md:27]: @_@ I swear I read the list like three times and didn't see that. Losing it.

File: truss/config.go [hunk 4]
```diff
@@ -1,35 +1,19 @@
...
-	"path/filepath"
-)
-
 // Config defines the inputs to a truss service generation
 type Config struct {
 	// The first path in $GOPATH
-	GOPATH string
+	GoPath []string
 
 	// The go packge where .pb.go files protoc-gen-go creates will be written
```
hasIan: "packge"

hasIan:
Looks good. Noticed a minor typo. :100: :shipit:

PR #93, opened 2016-11-15
Encode errors as json over http transport

hasAdamr:
[+1 line for actual change.](https://github.com/TuneLab/go-truss/compare/master...hasAdamr:encode-all-errors-json?expand=1#diff-0fe7c6aec1f529107babf5f63bddb542R234)

Plus test to make sure the change persists.

Errors now go across the wire as json with the key "`error`".

### DIFF
--- a/cmd/_integration-tests/transport/handlers/server/server_handler.go
+++ b/cmd/_integration-tests/transport/handlers/server/server_handler.go
@@ -4,6 +4,7 @@ package handler
 // implementation. It also includes service middlewares.
 
 import (
+	"fmt"
 	"golang.org/x/net/context"
 
 	pb "github.com/TuneLab/go-truss/cmd/_integration-tests/transport/transport-service"
@@ -62,4 +63,9 @@ func (s transportService) CtxToCtx(ctx context.Context, in *pb.MetaRequest) (*pb
 	}
 
 	return &resp, nil
+}
+
+// ErrorRPC implements Service.
+func (s transportService) ErrorRPC(ctx context.Context, in *pb.Empty) (*pb.Empty, error) {
+	return nil, fmt.Errorf("This error should be json over http transport")
 }
--- a/cmd/_integration-tests/transport/handlers/server/server_handler.go
+++ b/cmd/_integration-tests/transport/handlers/server/server_handler.go
@@ -4,7 +4,7 @@ package handler
 // implementation. It also includes service middlewares.
 
 import (
-	"fmt"
+	"errors"
 	"golang.org/x/net/context"
 
 	pb "github.com/TuneLab/go-truss/cmd/_integration-tests/transport/transport-service"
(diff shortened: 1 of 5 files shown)


cmd/_integration-tests/transport/handlers/server/server_handler.go
```diff
@@ -63,3 +64,8 @@ func (s transportService) CtxToCtx(ctx context.Context, in *pb.MetaRequest) (*pb
 
 	return &resp, nil
 }
+
+// ErrorRPC implements Service.
+func (s transportService) ErrorRPC(ctx context.Context, in *pb.Empty) (*pb.Empty, error) {
+	return nil, fmt.Errorf("This error should be json over http transport")
```
hasIan: This is a test, so, not a big deal, and this is probably known, but as a reminder: under normal circumstances, if an error is static, declare it with `errors.New` at the package level and just reference it here.

cmd/_integration-tests/transport/http_test.go
```diff
@@ -234,13 +231,34 @@ func TestCtxToCtxViaHTTPHeaderRequest(t *testing.T) {
...
-		t.Fatal(errors.Wrapf(err, "json error, got json: %q", string(respBytes)))
+		t.Fatal(errors.Wrapf(err, "json error, got response: %q", string(respBytes)))
 	}
 
 	if resp.V != value {
 		t.Fatalf("Expect: %q, got %q", value, resp.V)
 	}
+}
+
+func TestErrorRPCReturnsJsonError(t *testing.T) {
```
hasIan: Nit: be consistent with initialisms: `RPC` and `JSON`, not `RPC` and `Json`.

cmd/_integration-tests/transport/transport-test.proto
```diff
@@ -27,8 +27,15 @@ service TransportPermutations {
...
     };
   }
+  rpc ErrorRPC (Empty) returns (Empty) {
+    option (google.api.http) = {
+      get: "/error"
+    };
+  }
 }
 
+message Empty {}
```
hasIan, line 37: Trivia: Did you know that [Protobuf defines an `Empty` type](https://github.com/google/protobuf/blob/master/src/google/protobuf/empty.proto)? Doesn't really matter, but I have seen a bunch of empty redefinitions in our MV protobuf stuff.

gengokit/httptransport/templates.go
```diff
@@ -231,6 +231,7 @@ func MakeHTTPHandler(ctx context.Context, endpoints Endpoints, logger log.Logger
 	{{- if .HTTPHelper.Methods}}
 		serverOptions := []httptransport.ServerOption{
 			httptransport.ServerBefore(headersToContext),
+			httptransport.ServerErrorEncoder(errorEncoder),
```
hasIan, line 234: Here's where the magic happens! 💯 👍

hasIan (commented):
Looks good. Just minor comments, not blocking merge.

cmd/_integration-tests/transport/transport-test.proto
hasAdamr: In the current version of truss it is not possible to import that file. As when truss invokes `protoc` it does not import the `GOPATH` and such cannot find that file.

Though when pull #92 gets merged in, this file could be imported. Though the import line will be longer than the line that redefines `Empty`.

PR #95, opened 2016-11-16
Fix broken requests for paths with capital letters

hasLeland:
This is a fix for paths with capital letters not being properly handled by the server, with a test for this behavior. For example, if an rpc has an http binding to a path:

    /some/Path/With/Caps

The bug was that a client would make a request to the correct path `/some/Path/With/Caps`, but the server would register this method to a lowercased version of this path, `/some/path/with/caps`.

This is a fix for the above mis-behavior.

### DIFF
--- a/cmd/_integration-tests/transport/http_test.go
+++ b/cmd/_integration-tests/transport/http_test.go
@@ -42,6 +42,8 @@ func TestGetWithQueryClient(t *testing.T) {
 	}
 }
 
+// A manually-constructed HTTP request test, ensuring that a get with query
+// parameters works even outside the behvior of the client.
 func TestGetWithQueryRequest(t *testing.T) {
 	var resp pb.GetWithQueryResponse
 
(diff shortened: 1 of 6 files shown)


hasIan:
Dang, you beat me to it. :D Looks great though. Thanks. :100: :shipit:

cmd/_integration-tests/transport/http_test.go
```diff
@@ -239,6 +239,29 @@ func TestCtxToCtxViaHTTPHeaderRequest(t *testing.T) {
 	}
 }
 
+// Test that making a get request with capital letters in the path functions
+// correctly.
+func TestGetWithCapsPath(t *testing.T) {
```
hasAdamr: Could you add a test for a standard HTTP request as well? 

See [func TestGetWithQueryRequest](https://github.com/hasLeland/go-truss/blob/2bc8b2855686f329b05aa15675eb3026e2d0909c/cmd/_integration-tests/transport/http_test.go#L45) for an example

hasAdamr (approved):
Looks great!

Comment about adding a standard http test. Though that can come in on a different pull request.

hasLeland:
@hasAdamr Alright, I've implemented the manual request test. Should be it.

hasIan:
Why is `testHTTP` redeclared in every test function that wants to test using a raw HTTP request? It's parameterized and doesn't use closures, so it should be promoted to a package-level test function and just invoked in the particular test methods, rather than having it declared three (at least) times.

This isn't something that originated with this change, but with the additional raw HTTP request test that has been added, it's been exacerbated a bit, to the point where it makes sense to just change where it's declared, IMO.

It needn't be done prior to merging this, necessarily, but it should be done at some point.

hasAdamr:
@hasIan, it does use closures. `t *testing.T` and `var resp pb.GetWithQueryRequest` are closured into it. 

The reason why there cannot be one function for this is because we don't know what the response type will be. So we cannot unmarshal json into the the type unless it is closured in. That is why it is not in its own function.

Though the json could be unmarshalled into a `map[string]interface{}` and we know the keys so we could pull out `"V"` and typecast it to int to check the value.

The closure to get the `resp` type into the function and then copy and pasting seemed good at the time though.

They are for these kinds of tests: https://github.com/hasLeland/go-truss/blob/e6f3295013c5269d02eeb71817b4ff0b2f4b526b/cmd/_integration-tests/transport/http_test.go#L125

**UPDATE**: It seems we could use `"reflect"`? @hasLeland? I have not used reflect but it could be on the backlog to change this to a single function and use `"reflect"` to get the type.

hasIan:
OK, gotcha. Seems a little copy-pastey still. In these cases I use a higher-order function to return a function that closes over its parameters.

Because of the validity of the use case, let's just merge this and reduce SLOC count later.

PR #96, opened 2016-11-17
Update handler to not warn about params of unexported funcs

hasAdamr:
handler.go currently tries to update the params of unexported funcs. 

This was an oversight of mine that I noticed right after using truss, here is a quick fix.


hasLeland:
Lgtm

PR #98, opened 2016-11-22
modify template to fix problem with only parts of the generated code recognizing vendored go-kit type

tuneliza:


hasAdamr:
Could you provide an example of what this change accomplishes? It seems an import was removed, but an upcoming change I am working on for middlewares may put the `"github.com/go-kit/kit/endpoints"` import back into this file.

tuneliza:
I generated the truss service for tmcentities, so that the *pb.go files are in the  interface-defs/tmcentitiespb folder, and the rest of the service is in the services/tmcentities folder. When trying to build the service, I was getting this error x11:
```
./server_main.go:56: cannot use svc.MakeSiteGetByAdvertiserPackageEndpoint(service) (type "github.com/TuneLab/mv-universe-v2/vendor/github.com/go-kit/kit/endpoint".Endpoint) as type "github.com/go-kit/kit/endpoint".Endpoint in assignment
```
Zaq traced the problem to the template, adjusting the template solved the issue. I don't know why. (I have go-kit vendored because it was a dependency for something in bipipeline's glide file; we also tried to remove the vendored directory altogether, which only broke more things, so we put it back).

zaquestion:
@hasAdamr  I wouldn't expect the middleware changes to have any explicit type mention in this file. If its required I suppose we'll have to live with it and investigate why the vendored gokit isn't getting picked up in both cases. It does work fine if you remove the go-kit vendoring though thats not a reasonable expectation.

The simple solution is to just not explicitly call all the type. Which our upcomming injectMiddlewares func should be conformant too. Also this generates much prettier code.

hasAdamr:
@zaquestion  
> The simple solution is to just not explicitly call all the type.

Sorry, I'm having a hard time understanding this statement.

This change looks fine, just wanted to understand why it was pull requested in order for a test to be written or for code not to be introduced to cause the issue again.

@tuneliza, looks like this got out of sync with the other pull request that got merged in. 

I'll regenerated the `gengokit/template/template.go` on my local machine and merge this in.

hasAdamr:
I have now merged this pull request in. The manual regenerating of `gengokit/template/template.go` has made github not notice that this actually was merged in. 

So I am closing this, but the changes are in master.

Issue #41, opened 2016-09-09
Startup ports can only be configured via command line flags


zaquestion:
```
        debugAddr = flag.String("debug.addr", ":8080", "Debug and metrics listen address")
        httpAddr  = flag.String("http.addr", ":8081", "HTTP listen address")
        grpcAddr  = flag.String("grpc.addr", ":8082", "gRPC (HTTP) listen address")
```

It could be nice to allow at least the httpAddr to be specified with the PORT env variable. This is how heroku and dokku specify the port for the application and how they handle subdomain routing. Both heroku and dokku operate with procfiles so it may be possible to specify the port in the run command in there. I'll update here when I've tried.

zaquestion:
This passing the env in with the heroku/dokku Profile works.
https://github.com/zaquestion/current/blob/master/Procfile

```
web: current -http.addr :$PORT
```

zaquestion:
Closing out as this works fine. Not against revisiting in the future however.

PR #99, opened 2016-11-22
Add support for setting server ports via environment variables

hasLeland:
Adds ability to set the various ports of a truss-generated service by setting environment variables, as requested in #41. The mapping is as follows:

    Port of Transport   Environment variable to set
    -----------------   ---------------------------
    HTTP                TRUSS_HTTP_ADDR
    gRPC                TRUSS_GRPC_ADDR
    Debug port          TRUSS_DEBUG_ADDR

The format of each must be set with the '{INTERFACE}:{PORT}' format, same format as must be passed via command line arguments.


zaquestion:
I actually closed #41 as the command line args work fine, but happy to have these :)

What do you think about exposing `PORT` as an environment variable for HTTP, this would enable truss services to automagically work on heroku/ish environments.

I also question the TRUSS prefix here. Perhaps using the `NAME` would be more appropriate or just removing the prefix altogether. I'm not worried about collision as these should be specified in clean environments.

hasLeland:
@zaquestion I like the bare name, and I like the idea of setting the HTTP port with just 'PORT'. I'll make those changes.

hasLeland:
Merging because all tests pass and this now fulfills the enhancement requests given.

PR #100, opened 2016-11-23
Fix for ports not correctly handling null values

hasLeland:
The code I pushed for getting prots via environment variables had a bug in it, now fixed.


hasAdamr:
Seems ok, though I think we should just have the defaults as `const`'s and refer to them in two places.
```
const defaultDebug = ":5060"
const defaultHTTP = ":5050"
const defaultGRPC = ":5040"
```
```
debugAddr = flag.String("debug.addr", defaultDebug, ...
```
```
... && *debugAddr != defaultDebug {
```

tuneliza:
Merging per Zaq?, @hasLeland 👍  to suggest from @hasAdamr please follow up.

PR #101, opened 2016-11-28
Refactor names and tests in preperation for middleware addition

hasAdamr:
This refactor pull request is intended to separate out some reorganizing done to add middlewares. **No functional change.**

### Main points
- `gengokit.TemplateExecutor` renamed to `gengokit.Executor` as well as a rename to `gengokit.NewTemplateExecutor` to what you would expect. This creates a big part of the diff.
- `generateResponseFile` now takes the previous version of the file rather than a map of strings to all files, separating out the concern of finding getting the old file. 
- `ApplyTemplate` helper methods have been added to `gengokit` and package `handler` now uses these to apply its templates rather than every package reimplementing it. Package `middlewares` will also use this.
- Package `handler` now exports the file it renders as a constant and the `generator` package switches on that constant rather than it being defined in two places.

### DIFF
--- a/gengokit/generator/gen.go
+++ b/gengokit/generator/gen.go
@@ -2,13 +2,10 @@
 package generator
 
 import (
-	"bytes"
 	"go/format"
 	"io"
 	"io/ioutil"
-	"os"
 	"strings"
-	"text/template"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/pkg/errors"
@@ -21,32 +18,26 @@ import (
 	"github.com/TuneLab/go-truss/truss"
 )
 
-func init() {
-	log.SetLevel(log.InfoLevel)
-	log.SetOutput(os.Stderr)
-	log.SetFormatter(&log.TextFormatter{
-		ForceColors: true,
-	})
-}
-
 // GenerateGokit returns a gokit service generated from a service definition (svcdef),
 // the package to the root of the generated service goPackage, the package
 // to the .pb.go service struct files (goPBPackage) and any prevously generated files.
 func GenerateGokit(sd *svcdef.Svcdef, conf gengokit.Config) ([]truss.NamedReadWriter, error) {
-	te, err := gengokit.NewTemplateExecutor(sd, conf)
+	executor, err := gengokit.NewExecutor(sd, conf)
 	if err != nil {
 		return nil, errors.Wrap(err, "cannot create template executor")
 	}
 
-	fpm := make(map[string]io.Reader, len(conf.PreviousFiles))
+	prevFiles := make(map[string]io.Reader, len(conf.PreviousFiles))
 	for _, f := range conf.PreviousFiles {
-		fpm[f.Name()] = f
+		prevFiles[f.Name()] = f
 	}
 
 	var codeGenFiles []truss.NamedReadWriter
 
 	for _, templFP := range templFiles.AssetNames() {
-		file, err := generateResponseFile(templFP, te, fpm)
+		actualFP := templatePathToActual(templFP, sd.PkgName)
+		prev := prevFiles[actualFP]
+		file, err := generateResponseFile(executor, prev, templFP)
 		if err != nil {
 			return nil, errors.Wrap(err, "cannot render template")
 		}
@@ -62,32 +53,28 @@ func GenerateGokit(sd *svcdef.Svcdef, conf gengokit.Config) ([]truss.NamedReadWr
 
 // generateResponseFile contains logic to choose how to render a template file
 // based on path and if that file was generated previously. It accepts a
-// template path to render, a templateExecutor to apply to the template, and a
-// map of paths to files for the previous generation. It returns a
-// truss.NamedReadWriter representing the generated file
-func generateResponseFile(templFP string, te *gengokit.TemplateExecutor, prevGenMap map[string]io.Reader) (truss.NamedReadWriter, error) {
+// template path to render, a executor to apply to the template,
+// and . It returns a truss.NamedReadWriter representing the generated file.
+func generateResponseFile(executor *gengokit.Executor, file io.Reader, templFP string) (truss.NamedReadWriter, error) {
 	var genCode io.Reader
 	var err error
 
 	// Get the actual path to the file rather than the template file path
-	actualFP := templatePathToActual(templFP, te.PackageName)
-	// If we are rendering the server and or the client
-	if templFP == "NAME-service/handlers/server/server_handler.gotemplate" {
-		file := prevGenMap[actualFP]
-		h, err := handler.New(te.Service, file, te.PackageName)
+	actualFP := templatePathToActual(templFP, executor.PackageName)
+
+	switch templFP {
+	case handler.ServerFile:
+		h, err := handler.New(executor.Service, file, executor.PackageName)
 		if err != nil {
 			return nil, errors.Wrapf(err, "cannot parse previous handler: %q", actualFP)
 		}
 
-		if genCode, err = h.Render(templFP, te); err != nil {
-			return nil, errors.Wrap(err, "cannot render template")
+		if genCode, err = h.Render(templFP, executor); err != nil {
+			return nil, errors.Wrapf(err, "cannot render template: %s", templFP)
 		}
-	}
-
-	// if no code has been generated just apply the template
-	if genCode == nil {
-		if genCode, err = applyTemplateFromPath(templFP, te); err != nil {
-			return nil, errors.Wrap(err, "cannot render template")
+	default:
+		if genCode, err = applyTemplateFromPath(templFP, executor); err != nil {
+			return nil, errors.Wrapf(err, "cannot render template: %s", templFP)
 		}
 	}
 
@@ -125,30 +112,13 @@ func templatePathToActual(templFilePath, packageName string) string {
 }
 
 // applyTemplateFromPath calls applyTemplate with the template at templFilePath
-func applyTemplateFromPath(templFilePath string, executor *gengokit.TemplateExecutor) (io.Reader, error) {
-	templBytes, err := templFiles.Asset(templFilePath)
+func applyTemplateFromPath(templFP string, executor *gengokit.Executor) (io.Reader, error) {
+	templBytes, err := templFiles.Asset(templFP)
 	if err != nil {
-		return nil, errors.Wrapf(err, "unable to find template file: %v", templFilePath)
+		return nil, errors.Wrapf(err, "unable to find template file: %v", templFP)
 	}
 
-	return applyTemplate(templBytes, templFilePath, executor)
-}
-
-func applyTemplate(templBytes []byte, templName string, executor *gengokit.TemplateExecutor) (io.Reader, error) {
-	templateString := string(templBytes)
-
-	codeTemplate, err := template.New(templName).Funcs(executor.FuncMap).Parse(templateString)
-	if err != nil {
-		return nil, errors.Wrap(err, "cannot create template")
-	}
-
-	outputBuffer := bytes.NewBuffer(nil)
-	err = codeTemplate.Execute(outputBuffer, executor)
-	if err != nil {
-		return nil, errors.Wrap(err, "template error")
-	}
-
-	return outputBuffer, nil
+	return executor.ApplyTemplate(string(templBytes), templFP)
 }
 
 // formatCode takes a string representing golang code and attempts to return a
@@ -164,5 +134,4 @@ func formatCode(code []byte) []byte {
 	}
 
 	return formatted
-
 }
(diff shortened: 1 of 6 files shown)


gengokit/generator/gen.go
```diff
@@ -62,32 +53,28 @@ func GenerateGokit(sd *svcdef.Svcdef, conf gengokit.Config) ([]truss.NamedReadWr
...
-	if templFP == "NAME-service/handlers/server/server_handler.gotemplate" {
-		file := prevGenMap[actualFP]
-		h, err := handler.New(te.Service, file, te.PackageName)
+	actualFP := templatePathToActual(templFP, executor.PackageName)
+
+	switch templFP {
+	case handler.ServerFile:
+		h, err := handler.New(executor.Service, file, executor.PackageName)
 		if err != nil {
 			return nil, errors.Wrapf(err, "cannot parse previous handler: %q", actualFP)
```
hasIan, line 69: This error doesn't make as much sense when reading this code, now that there's no mention of "previous files" here. There's context in the error message that isn't present in the actual code.

gengokit/generator/gen.go
```diff
@@ -62,32 +53,28 @@ func GenerateGokit(sd *svcdef.Svcdef, conf gengokit.Config) ([]truss.NamedReadWr
 
 // generateResponseFile contains logic to choose how to render a template file
 // based on path and if that file was generated previously. It accepts a
-// template path to render, a templateExecutor to apply to the template, and a
-// map of paths to files for the previous generation. It returns a
-// truss.NamedReadWriter representing the generated file
-func generateResponseFile(templFP string, te *gengokit.TemplateExecutor, prevGenMap map[string]io.Reader) (truss.NamedReadWriter, error) {
+// template path to render, a executor to apply to the template,
+// and . It returns a truss.NamedReadWriter representing the generated file.
+func generateResponseFile(executor *gengokit.Executor, file io.Reader, templFP string) (truss.NamedReadWriter, error) {
```
hasIan: I don't like `generateResponseFile` too much as the name of this function. I'm not sure where the idea of a "response" ties in here. Part of the issue is that it's doing a few things. When it's hard to succinctly and accurately describe what a function does, it's a sign that it may be doing too much.

Reading this function, it gets the name of a generated file, tests to see if it's `handler.ServerFile` (`handler` itself seems like an inadequately descriptive package name here). If it's this file, it reads it and updates it; if it's not, it renders a new file from a template. Then it immediately reads the code that has been generated, formats it, writes it again, and returns this writer to the caller.

It's no wonder that it was difficult to name.

gengokit/generator/gen_test.go
```diff
@@ -116,7 +115,7 @@ func svcMethodsNames(methods []*svcdef.ServiceMethod) []string {
 	return mNames
 }
 
-func stringToTemplateExector(def, importPath string) (*gengokit.TemplateExecutor, error) {
+func stringToTemplateExector(def, importPath string) (*gengokit.Executor, error) {
```
hasIan: "Exector"

I also find `Executor` kind of a strange noun for this type, since it's not actually doing the execution. Given how this type gets passed as a value to an "apply template" function, It seems more like a `Values` than `Executor`.

gengokit/generator/gen_test.go
```diff
@@ -217,64 +220,55 @@ func TestAllTemplates(t *testing.T) {
...
-			continue
-		}
-		prevGenMap := make(map[string]io.Reader)
+		var prev io.Reader
 
-		firstCode, err := testGenerateResponseFile(templFP, te, prevGenMap)
+		firstCode, err := testGenerateResponseFile(executor1, prev, templFP)
 		if err != nil {
-			t.Fatalf("%v failed to format on first generation\n\nERROR:\n\n%v\n\nCODE:\n\n%v", templFP, err.Error(), string(firstCode))
+			t.Fatalf("%s failed to format on first generation\n\nERROR:\n\n%s\n\nCODE:\n\n%s", templFP, err.Error(), firstCode)
```
hasIan: Small tip: you don't need to call `Error()` on errors. `fmt` and `testing` knows how to pull a string representation from an `error` type given a `%s` or a `%v`.

gengokit/generator/gen_test.go
```diff
@@ -346,82 +340,79 @@ func TestUpdateMethods(t *testing.T) {
...
+	thirdCode, err := renderService(svc, secondCode, te, templPath)
+	if err != nil {
+		t.Fatal(err)
+	}
 
-		forthCode, err := renderService(svc, thirdCode, te, templPath)
-		if err != nil {
-			t.Fatal(err)
-		}
+	if thirdCode <= secondCode {
```
hasIan: This will practically never affect this code path, but this test is not a reliable method for determining the length of a string, [as illustrated here](https://play.golang.org/p/Tm45ji81iv). If you want to test the length of a string, use `len(str)`.

gengokit/generator/gen_test.go
```diff
@@ -346,82 +340,79 @@ func TestUpdateMethods(t *testing.T) {
...
+	forthCode, err := renderService(svc, thirdCode, te, templPath)
+	if err != nil {
+		t.Fatal(err)
+	}
 
-		fifthCode, err := renderService(svc, forthCode, te, templPath)
-		if err != nil {
-			t.Fatal(err)
-		}
+	if forthCode >= thirdCode {
```
hasIan: "fourth" not "forth"

gengokit/generator/gen_test.go
```diff
@@ -346,82 +340,79 @@ func TestUpdateMethods(t *testing.T) {
...
-	return gentesthelper.DiffStrings(
-		a,
-		b,
-	)
+	if fifthCode <= forthCode {
+		t.Fatal("Generated code not longer after regenerated with additional service method\n" +
+			templPath + "\n" +
+			diff(forthCode, fifthCode))
+	}
 }
```
hasIan, line 406: I see what this test is doing, and that's cool, but it kinda makes me @_@. 😄

gengokit/generator/gen_test.go
```diff
@@ -346,82 +340,79 @@ func TestUpdateMethods(t *testing.T) {
...
 
-func renderService(svc *svcdef.Service, prev string, te *gengokit.TemplateExecutor, templPath string) (string, error) {
-	h, err := handler.New(svc, strings.NewReader(prev), te.PackageName)
+func renderService(svc *svcdef.Service, prev string, te *gengokit.Executor, templPath string) (string, error) {
+	var prevFile io.Reader
```
hasIan, line 409: This isn't a file at all, actually.

gengokit/generator/gen_test.go
```diff
@@ -436,46 +427,53 @@ func renderService(svc *svcdef.Service, prev string, te *gengokit.TemplateExecut
...
 
-func testGenerateResponseFile(templFP string, te *gengokit.TemplateExecutor, prevGenMap map[string]io.Reader) ([]byte, error) {
+func diff(a, b string) string {
+	return gentesthelper.DiffStrings(
+		a,
+		b,
+	)
+}
+
+func testGenerateResponseFile(executor *gengokit.Executor, prev io.Reader, templFP string) (string, error) {
```
hasIan: This is a bit of a nit, but as a reader it tripped me up ever so slightly. This is a little weird because it's a helper function but its return type is not the same as `generateResponseFile`, which returns a `NamedReader` or something like that. This sort of creates some dissonance between the implementation and the test, given the similarity of the names. Perhaps give the test function a distinct name. Of course, I wasn't crazy about `generateResponseFile` to begin with, but the point stands regardless of the name of the non-test function.

Further, this test function appears to be doing redundant work--it reads the generated code, then formats it. This happens within `generateResponseFile` though. If there's a reason for it being done again here, it's not obvious without really digging in.

gengokit/gengokit.go
```diff
@@ -56,3 +58,25 @@ func NewTemplateExecutor(sd *svcdef.Svcdef, conf Config) (*TemplateExecutor, err
 		FuncMap:      funcMap,
 	}, nil
 }
+
+// ApplyTemplate applies the passed template with the Executor
```
hasIan: This documentation comment doesn't really tell me very much. If this were called `Render`, I could document it like, "Render returns a template filled with the Executor field values."

This is another case where `Executor` seems like a weird name. And this rendering function being a method on a set of template values strikes me as a little weird. Like, this would make the most sense as a function that takes a template reference and a set of values that should be available to the template.

That's a nitpick, but as it is, it seems like functionality bolted onto a somewhat strange location.

gengokit/handler/handler.go
```diff
@@ -23,7 +23,9 @@ import (
 // NewService is an exported func that creates a new service
 // it will not be defined in the service definition but is required
 const ignoredFunc = "NewService"
-const serverTemplPath = "NAME-service/handlers/server/server_handler.gotemplate"
+
+// ServerFile is the path to the service handelr file that handler will render
```
hasIan: "handelr"

It's a mouthful of a sentence anyway, though. How about just "ServerFile is the path to the service handler template file."

It'd probably be better named `ServerHandlerTemplate` or similar. More explicit and more accurate.

gengokit/handler/handler.go
```diff
@@ -283,30 +285,15 @@ func exprString(e ast.Expr) string {
...
 }
 
 func applyServerMethsTempl(exec handlerExecutor) (io.Reader, error) {
-	return applyTemplate(serverMethsTempl, "ServerMethsTempl", exec)
-}
-
-func applyTemplate(templ string, templName string, exec interface{}) (io.Reader, error) {
 	funcMap := template.FuncMap{
 		"ToLower": strings.ToLower,
 		"GoName":  generatego.CamelCase,
```
hasIan: One thing I notice is that the same `FuncMap` is set up in a couple places, and it's the same everywhere, except in one place an additional "Titles" function is set up. Why not just declare this `FuncMap` as a package level variable rather than redefining it each time templates are going to be rendered?

hasIan (changes requested):
Looks good. I've addressed some things that aren't necessarily a direct result of the changes here, but more things that have been lying dormant. There are some nits, some maybe worth addressing. But pretty minor. A little bit of polish and this is good to go.

gengokit/generator/gen_test.go
hasAdamr: I don't any attachment to the name, I thought I was following a pattern in the standard library calling it Executor but I was mistaken.

The [`text/templates`](https://golang.org/pkg/text/template/#Template.Execute) calls this value the `data`.  

```
func (t *Template) Execute(wr io.Writer, data interface{}) error
```

So my vote would be for `gengokit.Data`, but if you feel strongly about `gengokit.Values` I would be fine with that too.

gengokit/generator/gen.go
hasAdamr: `file` should probably be named `prevFile`. Then that context should be understandable.

gengokit/handler/handler.go
hasAdamr: I went with `ServerHandlerPath`.

gengokit/generator/gen_test.go
hasAdamr: This function is doing the work of getting the `string` out of the `truss.NamedReadWriter` (note that there is a pull request up right now to remove the `truss.NamedReadWriter` interface.

Also generateResponseFile will try and format the output, but it may fail. Though it always returns code. This is useful for manual debugging because truss will always write code to disk, even if it does not format or compile. 

This function on the other had, will return an error if it cannot format, and this is used in the tests to find which files do not format.

gengokit/generator/gen_test.go
hasAdamr: It represents the prevous version of a file, that is what this helper function is for.

In a pull #104 this function moves to the `handler` package and I can update the pull request to have a better comment for this helper function

gengokit/generator/gen.go
hasAdamr: This area is being refactored currently, but Zaq wanted middlewares in and so I have done this minimal refactor in order to get the middlewares in. 

So basically, I hear all of your points, and I have branches with the code refactored. But it is too large of a change to get into the repo in a reasonable time frame. So your points will be addressed in another refactor.

hasIan: Sounds good.

gengokit/generator/gen_test.go
hasIan: `Data` works. I generally have a distaste for this as an identifier because of its extreme overuse in some of the legacy code here, all over the place, where its use does not communicate any meaning at all. As long as its use is restricted to narrowly-scoped cases, it's fine.

gengokit/generator/gen_test.go
```diff
@@ -29,6 +28,7 @@ func init() {
 
 func init() {
 	log.SetLevel(log.DebugLevel)
+	_ = errors.Wrap
```
hasIan: If this is the only place that `errors` is referenced, we should just remove the import. If this isn't the only place that `errors` is referenced, we should remove this line.

hasIan:
Looks good, everything else can be addressed as we evolve things. I did add one more comment about a `_ = errors.Wrap` line in a test file and whether that's necessary. Not an issue for this PR, just a note on something that may be unnecessary.

hasIan (approved):

Issue #94, opened 2016-11-15
Proposal: Truss service middlewares


hasAdamr:
Overview: 
1. Have a function that truss does not overwrite where middlewares are added `injectServiceMiddlewares `.
2. Provide options to generate middlewares via flags.
3. User hooks up middlewares in desired order in `injectServiceMiddlewares`.

This uses the `example/echo.proto` for types.

1. Have `echo-server/server_main.go` call a `injectServiceMiddlewares` function to add middlewares
 
	```golang
	// echo-service/echo-server/server_main.go
	...
	// Business domain.
	var service pb.EchoServer
	{
		service = handler.NewService()
		service = injectServiceMiddlewares(service)
	}
	```
	```
	// echo-service/echo-server/wrap.go, next to server_main.go?
	package main
	
	import (
		pb "github.com/TuneLab/go-truss/example/echo-service"
	)
	
	func injectServiceMiddlewares(in pb.EchoServer) pb.EchoServer {
		// TODO: Add middlewares
		return in
	}
	```

2. Generate a empty middleware with truss.
	```
	truss -sm logger [*.proto]
	// or
	truss --servce-middleware logger [*.proto]
	```
	To create:
	```
	// echo-service/middlewares/logger/logger.go
	package logger 
	import (
		"golang.org/x/net/context"
	
		pb "github.com/TuneLab/go-truss/example/echo-service"
	)
	
	type Middleware func(pb.EchoServer) pb.EchoServer
	
	type serviceLogger struct {
		next pb.EchoServer
		// TODO: Add serviceLogger fields 
	}
	
	func (s *serviceLogger) Echo(ctx context.Context, in *pb.EchoRequest) (*pb.EchoResponse, error) {
		// TODO: Implement Echo for serviceLogger
		return s.next.Echo(ctx, in)
	}

	func (s *serviceLogger) Louder(ctx context.Context, in *pb.LouderRequest) (*pb.EchoResponse, error) {
		// TODO: Implement Louder for serviceLogger
		return s.next.Louder(ctx, in)
	}

	func NewServiceLogger( /* TODO: Add serviceLogger constructor params */ ) Middleware {
		return func(next pb.EchoServer) pb.EchoServer {
			return &serviceLogger{
				next: next,
				// TODO: Add serviceLogger constructor params
			}
		}
	}
	```

3. Hooking it up in `wrap.go`
	```
	package main
	
	import (
		pb "github.com/TuneLab/go-truss/example/echo-service"

[...]

PR #102, opened 2016-11-28
Endpoints and Service middleware injector templates and generator.

hasAdamr:
This pull request adds code to hook up middlewares in the generated code output of truss. Implemented close to issue #94, details below.

This pull adds two files to the generator truss service.
```
NAME-service/middlewares/endpoints.go
NAME-service/middlewares/service.go
```

In these files are the `func InjectEndpointsMiddlewares` and `func InjectServiceMiddlewares` funcs described in issue #94. The functions are hooked up to `func main()` as described in the issue as well.

Currently, the gengokit package `middlewares` (which renders the middlewares) will only ever create files if they are not there and will otherwise not modify them. Though this can be updated in the future to inject additional functions.


gengokit/generator/gen.go
```diff
@@ -76,6 +77,18 @@ func generateResponseFile(data *gengokit.Data, prevFile io.Reader, templFP strin
 		if genCode, err = applyTemplateFromPath(templFP, data); err != nil {
 			return nil, errors.Wrapf(err, "cannot render template: %s", templFP)
 		}
+	case middlewares.EndpointsPath:
```
hasIan: Keep `default` as the last case in `switch` blocks, just so it reads nicely.

gengokit/middlewares/middlewares.go
```diff
@@ -0,0 +1,63 @@
...
+
+// EndpointsPath is the path to the Endpoints middleware file that package middlewares renders
+const EndpointsPath = "NAME-service/middlewares/endpoints.gotemplate"
+
+// ServicePath is the path to the Service middleware file that package middlewares renders
+const ServicePath = "NAME-service/middlewares/service.gotemplate"
+
+// New returns a Middleware which can render EndpointsFile and ServiceFile as
+// well as read in previous versions of each respective file
+func New() *Middlewares {
```
hasIan, line 21: This seems kind of pointless unless the function actually does something. This could be replaced with

```
foo := new(middlewares.Middlewares)
```

gengokit/middlewares/middlewares.go
```diff
@@ -0,0 +1,63 @@
...
+
+// LoadService loads a previous version of ServiceFile
+func (m *Middlewares) LoadService(prev io.Reader) {
+	m.prevService = prev
+}
+
+// Render can render either EndpointsPath or ServicePath. With no previous
+// version it renders the templates, if there was a previous version loaded in,
+// it passes that through
+func (m *Middlewares) Render(alias string, data *gengokit.Data) (io.Reader, error) {
```
hasIan: It's not clear to me how this parameter is an alias. Isn't this just a template file?

gengokit/middlewares/middlewares_test.go
```diff
@@ -0,0 +1,236 @@
...
+		t.Fatal(err)
+	}
+
+	wantFormatted, serviceFormatted, diff := thelper.DiffGoCode(wantService, string(serviceBytes))
+	if wantFormatted != serviceFormatted {
+		t.Fatalf("Serivce middleware different than expected:\n\n%s", diff)
+	}
+}
+
+func TestNewEndpoitnMiddleware(t *testing.T) {
```
hasIan: "Endpoitn"

gengokit/middlewares/middlewares_test.go
```diff
@@ -0,0 +1,236 @@
...
+		t.Fail()
+	}
+
+	middleware := New()
+
+	middleware.LoadService(strings.NewReader(wantService))
+
+	service, err := middleware.Render(ServicePath, data)
+	if err != nil {
+		t.Fail()
```
hasIan: Is `Fail` actually what you want here? Mark the test as failed but continue execution? Seems like an error like this would prevent any further execution of the test from being worthwhile.

gengokit/middlewares/middlewares_test.go
```diff
@@ -0,0 +1,236 @@
...
+	_, data, err := generalService()
+	if err != nil {
+		t.Fail()
+	}
+
+	middleware := New()
+
+	_, err = middleware.Render("not/valid/file.go", data)
+	if err == nil {
+		t.Fatalf("Tried to render unknown file")
```
hasIan: This doesn't mean much to a non-author. Add "this should have produced an error, but didn't" or something like that.

gengokit/template/NAME-service/NAME-server/server_main.gotemplate
```diff
@@ -60,19 +60,21 @@ func main() {
...
-		// add service logging and metrics here
+		// Inject Service middlewares. See ../middlewares/service.go
+		service = middlewares.InjectServiceMiddlewares(service)
 	}
 
 	// Endpoint domain.
 	var (
 	{{range $i := .Service.Methods -}}
-		{{ToLower $i.Name}}Endpoint = svc.Make{{$i.Name}}Endpoint(service)
+			{{ToLower $i.Name}}Endpoint = svc.Make{{$i.Name}}Endpoint(service)
```
hasIan: Looks like a rogue tab got in here. Figuring that gofmt won't work on these files since they're not actually valid Go, so we need to watch formatting closely ourselves in the templates.

hasIan (commented):
Looks good, just some minor comments.

gengokit/generator/gen.go
hasAdamr: This must have happened in the rebase and I did not notice. Woops, thanks for the catch.

gengokit/middlewares/middlewares.go
hasAdamr: So I am not attached to this method though here is my reasoning. Please tell me what you think.

I added this method to be symetrical with the `handler` package. Though at this point, it does not take any parameters, does no setup, and cannot fail. This is because the middlewares are not doing anything complicated.

But new features may add requirements for setting things up. (e.g. updating middlewares based on the .proto file, inserting middlewares). So doing a `new(middlewares.Middlewares)` and then just calling `Load...` or `Render` may be a problem. 

Though this may be a case of wanting to add a wiggle when I don't need it.

If you think my reasoning is not up to snuff, I can change it to `new(middlewares.Middlewares)` and update it later to a `New` method if setup is required.

gengokit/middlewares/middlewares.go
hasAdamr: It is an alias to a file on disk. It is not the actual path to a file on disk as it will be in the `NAME-` format. But it is not a template file. It is a alias to path, where the `generator` package will update this alias to an actual path.

gengokit/middlewares/middlewares_test.go
hasAdamr: Woops, mistake.

gengokit/middlewares/middlewares.go
hasIan: If you think that we are going to have functionality that will warrant a `New` method to conceal setup in the `Middlewares` type, then I think it's fine to leave it. I hadn't considered the API symmetry with `handler`, and I don't really know how important it is, but it's not hurting anything, so we can proceed.

gengokit/middlewares/middlewares.go
hasIan: Hmm, okay. I understand what it is, but still think that "alias" is not an intuitive name here, especially given the explanation that is required to clarify it.

In general, aspire to create that which explains itself, rather than requiring another to explain it. This is always easier said than done, and in some cases, not practical. In this case, because we're just talking about a function parameter with limited scope, it's fine to proceed. I get hung up on naming because it's one of the hardest yet one of the most important things we do.

hasAdamr: Based on that I will update it to `path`. :)

gengokit/middlewares/middlewares.go
hasAdamr: I think that functionality will be coming soonish. I will leave it for now.

PR #103, opened 2016-11-28
Refactor to remove truss.NamedReadWriter complexity

hasAdamr:
This pull removes the `truss.NamedReadWriter` interface and `truss.SimpleFile` struct. **No functional change**

This interface was originally because we hoped that `os.File` would satisfy this interface, though the `Name()` method of `os.File` would not work correctly.

It is replaced with a `map[string]io.Reader` which represents a relative path on disk to the service directory to file contents. It is used as the structure for both reading in files and how code generating packages output files to write. 

It is overall cleaner as it removes a custom type and I hope reflects a slightly better understanding of `io.Reader`'s.

### DIFF
--- a/cmd/truss/main.go
+++ b/cmd/truss/main.go
@@ -242,7 +242,7 @@ func generateCode(cfg *truss.Config, dt deftree.Deftree, sd *svcdef.Svcdef) (map
 	return combineFiles(genGokitFiles, genDocFiles), nil
 }
 
-// combineFiles takes any numbers of map[string]io.Reader and combines them into one.
+// combineFiles takes any number of map[string]io.Reader and combines them into one.
 func combineFiles(group ...map[string]io.Reader) map[string]io.Reader {
 	final := make(map[string]io.Reader)
 	for _, g := range group {
@@ -322,45 +322,36 @@ func readPreviousGeneration(serviceDir string) (map[string]io.Reader, error) {
 	}
 
 	dir, _ := filepath.Split(serviceDir)
-	sfs := simpleFileConstructor{
-		dir:   dir,
-		files: make(map[string]io.Reader),
-	}
-	err := filepath.Walk(serviceDir, sfs.makeSimpleFile)
-	if err != nil {
-		return nil, errors.Wrapf(err, "cannot fully walk directory %v", sfs.dir)
-	}
+	files := make(map[string]io.Reader)
 
-	return sfs.files, nil
-}
+	addFileToFiles := func(path string, info os.FileInfo, err error) error {
+		if info.IsDir() {
+			return nil
+		}
 
-// simpleFileConstructor has function makeSimpleFile of type filepath.WalkFunc
-// This allows for filepath.Walk to be called with makeSimpleFile and build a
-// map[string]io.Reader for all files in a directory
-type simpleFileConstructor struct {
-	dir   string
-	files map[string]io.Reader
-}
+		file, ioErr := os.Open(path)
 
-// makeSimpleFile is of type filepath.WalkFunc
-// makeSimpleFile adds files as io.Readers to sfs.files by path
-func (sfs *simpleFileConstructor) makeSimpleFile(path string, info os.FileInfo, err error) error {
-	if info.IsDir() {
+		if ioErr != nil {
+			return errors.Wrapf(ioErr, "cannot read file: %v", path)
+		}
+
+		// trim the prefix of the path to the proto files from the full path to the file
+		relPath, err := filepath.Rel(dir, path)
+		if err != nil {
+			return err
+		}
+
+		files[relPath] = file
+
 		return nil
 	}
 
-	file, ioErr := os.Open(path)
-
-	if ioErr != nil {
-		return errors.Wrapf(ioErr, "cannot read file: %v", path)
+	err := filepath.Walk(serviceDir, addFileToFiles)
+	if err != nil {
+		return nil, errors.Wrapf(err, "cannot fully walk directory %v", dir)
 	}
 
-	// trim the prefix of the path to the proto files from the full path to the file
-	name := strings.TrimPrefix(path, sfs.dir)
-
-	sfs.files[name] = file
-
-	return nil
+	return files, nil
 }
 
 // fileExists checks if a file at the given path exists. Returns true if the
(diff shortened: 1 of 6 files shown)


cmd/truss/main.go
```diff
@@ -56,14 +55,18 @@ func main() {
...
+	for path, file := range genFiles {
+		err := writeGenFile(file, path, cfg.ServicePath)
 		if err != nil {
 			exitIfError(errors.Wrap(err, "cannot to write output"))
 		}
 	}
 }
 
+// fileSystem is a map of paths relative to truss.Config.ServicePath mapped to
+// io.Reader's as files
```
hasIan: `io.Readers`. Apostrophes are always, always possessive*, never, ever pluralizing. 

*exception: "it's," meaning "it is."

More semantically, the sentence is confusing. "map of X mapped to Y." This is an internal type that's clearly a map of strings to io.Readers. This type is internal, so documentation on what's pretty clear seems kind of unnecessary.

Further, though, this type has no methods attached, so I kind of think it'd make sense to get rid of it and just use `map[string]io.Reader` in its place, which is immediately understandable and doesn't further overload the word "filesystem," which doesn't seem like the most accurate name to me for this type.

cmd/truss/main.go
```diff
@@ -325,33 +335,30 @@ func readPreviousGeneration(serviceDir string) ([]truss.NamedReadWriter, error)
...
 
 // simpleFileConstructor has function makeSimpleFile of type filepath.WalkFunc
-// This allows for filepath.Walk to be called with makeSimpleFile and build a truss.SimpleFile
-// for all files in a direcotry
+// This allows for filepath.Walk to be called with makeSimpleFile and build a fileSystem
+// for all files in a directory
 type simpleFileConstructor struct {
 	dir   string
-	files []truss.NamedReadWriter
+	files map[string]io.Reader
```
hasIan: Here you've used the underlying definition of a `fileSystem` and I don't think anything's suffered for it.

cmd/truss/main.go
```diff
@@ -325,33 +335,30 @@ func readPreviousGeneration(serviceDir string) ([]truss.NamedReadWriter, error)
 }
 
 // simpleFileConstructor has function makeSimpleFile of type filepath.WalkFunc
-// This allows for filepath.Walk to be called with makeSimpleFile and build a truss.SimpleFile
-// for all files in a direcotry
+// This allows for filepath.Walk to be called with makeSimpleFile and build a fileSystem
+// for all files in a directory
 type simpleFileConstructor struct {
```
hasIan: I know this change wasn't introduced here, but seeing the word "constructor" in Go code kind of sets off my smell alarm. I know it doesn't really imply anything here, but since it doesn't, it shouldn't use a word that carries all that baggage with it.

gengokit/generator/gen.go
```diff
@@ -15,47 +16,38 @@ import (
...
 }
 
 // generateResponseFile contains logic to choose how to render a template file
 // based on path and if that file was generated previously. It accepts a
-// template path to render, a data to apply to the template,
-// and . It returns a truss.NamedReadWriter representing the generated file.
-func generateResponseFile(data *gengokit.Data, prevFile io.Reader, templFP string) (truss.NamedReadWriter, error) {
+// template path to render, a templateExecutor to apply to the template, and a
+// map of paths to files for the previous generation. It returns a
+// io.Reader representing the generated file
```
hasIan: Punctuate your sentences, yo. ;)

cmd/truss/main.go
```diff
@@ -240,33 +243,41 @@ func generateCode(cfg *truss.Config, dt deftree.Deftree, sd *svcdef.Svcdef) ([]t
 
 	genDocFiles := gendoc.GenerateDocs(dt)
 
-	genFiles := append(genGokitFiles, genDocFiles...)
+	return combineFiles(genGokitFiles, genDocFiles), nil
+}
+
+// combineFiles takes any numbers of fileSystems and combines them into one.
+func combineFiles(group ...fileSystem) fileSystem {
```
hasIan: Will there ever be a case where there can be a conflict? This function silently overwrites an existing key.

hasIan (commented):
Looks good. Simplification and expunging unnecessary code is always a nobel pursuit. Minor comments placed.

cmd/truss/main.go
hasAdamr: I thought that labeling it `fileSystem` would make functions that take in and return this value easier to read than:
```
func combineFiles(group ...fileSystem) fileSystem {
...
}

func combineFiles(group ...map[string]io.Reader) map[string]io.Reader {
...
}
```

If you disagree, then I have no attachment here.

cmd/truss/main.go
hasAdamr: It kind of is a constructor. The way I have seen recommended for walking a
filepath is to use `path/filepath.Walk` with a `path/filepath.WalkFunc`.

So `simpleFileConstructor` has a `WalkFunc` as a method, and that method
modifies the state of `simpleFileConstructor` and in the end the field `files`
of `simpleFileConstructor` are the files that were found. So it is sort of a
constructor.

Now that I know more, I can see I could use a function closure to achieve this
effect and the next time this code is touched that can be done.

cmd/truss/main.go
hasAdamr: It is possible.

Before truss was just writing from a slice, so items later in the slice would overwrite earlier items if they had the same path. Same issue, just clearer in this case.

I believe even if there was a conflict, truss should write what it can.

So I am not sure how to handle this. Options:
- Add check and just log.Warn within the function if this happens
- Return the output and an error if there was a conflict. Like `io.Reader` returns the number of bytes read and an errror. Then log after the function is called if there was an error

cmd/truss/main.go
hasIan: It's fewer characters, but it obfuscates its meaning. I have to know what a `fileSystem` is to read that function signature, whereas there's no ambiguity about what a `map[string]io.Reader` is. The Go way values explicitness and readability--where readability here is the ability to read the code and understand it without having to cross reference other things.

If you want to name the type, I would make it more plain rather than trying to abstract away its definition in the name. E.g., `readerMap`. But honestly, I feel like we don't need the type. It's not used so prolifically that the slight keystroke savings is worth the slight obfuscation.

PR #104, opened 2016-11-28
Move handler test

hasAdamr:
**No functional change**

Moving the updating methods test from the generator to the handler package. The generator originally handled this operation so the test was there but it makes more sense in the handler package.

This pull will collapse when #101 gets merged in.


gengokit/handler/handler_test.go
```diff
@@ -1,29 +1,34 @@
...
+var diff = helper.DiffStrings
+var testFormat = helper.TestFormat
 
 func init() {
 	gopath = filepath.SplitList(os.Getenv("GOPATH"))
 }
 
 func init() {
-	_ = thelper.DiffStrings
+	_ = helper.DiffStrings
```
hasIan: Why is this in here? The package is being used. This line should probably be deleted.

hasIan (commented):
Looks straightfoward. :shipit: One comment about what looks like an unnecessary dummy assignment.

PR #105, opened 2016-11-28
Move example to _example for tests

hasAdamr:
If `./example/` has code non compilable code in it from a truss run, then
`go test ./...` will exit non 0 from this code failing to build. This is
annoying when testing truss on the example and then trying to rerun the
tests. This fixes it by adding the _ to avoid go's tooling.


hasIan:
LGTM :shipit:

PR #108, opened 2016-12-06
Brevity in readme and update make dependencies target

zaquestion:
thoughts?

Make sure to click "rich diff" on the README

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,15 @@
 # Build native Truss by default.
 default: truss
 
-dependencies: 
-	go install github.com/golang/protobuf/protoc-gen-go
+dependencies:
+	go get github.com/go-kit/kit
+	go get google.golang.org/genproto
+	go get github.com/golang/protobuf/{proto,protoc-gen-go}
+
+update-dependencies:
+	go get -u github.com/go-kit/kit
+	go get -u google.golang.org/genproto
+	go get -u github.com/golang/protobuf/{proto,protoc-gen-go}
 
 # Generate go files containing the all template files in []byte form
 gobindata:
@@ -27,3 +34,5 @@ test-integration:
 # Removes generated code from tests
 testclean:
 	$(MAKE) -C cmd/_integration-tests clean
+
+.PHONY: testclean test-integration test-go test truss gobindata dependencies
--- a/README.md
+++ b/README.md
@@ -17,28 +17,21 @@ source.
 
 To install this software, you must:
 
-1. Install protocol buffers version 3 or greater. The easiest way is to
-download a release from the [github releases](https://github.com/google/protobuf/releases)
-and add the binary to your `$PATH`. 
-Otherwise [install from source.](https://github.com/google/protobuf/releases)
-2. Of course, install the Go compiler and tools from https://golang.org/.
-See https://golang.org/doc/install for details.
-3. Install the `protoc-gen-go` and `proto` packages for Go. The simplest 
-way is to run 
+1. Install protoc 3 or newer. The easiest way is to
+download a release from [github](https://github.com/google/protobuf/releases)
+and add to `$PATH`.
+Otherwise [install from source.](https://github.com/google/protobuf)
+1. Install the `proto` and `protoc-gen-go` packages. (`protoc-gen-go` must be in `$PATH`)
 
 	```
 	go get -u github.com/golang/protobuf/{proto,protoc-gen-go}
 	```
-
-	The compiler plugin, protoc-gen-go, will be installed in `$GOPATH/bin`. 
-It must be in your `$PATH` for the protocol compiler, `protoc`, to find it.
-4. Install google's protobuf/grpc http annotations option. `protoc` will use 
-these to compile http annotations. To install run:
+1. Install HTTP Annotations
 
 	```
-	go get -u google.golang.org/genproto	
+	go get -u google.golang.org/genproto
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -27,3 +28,5 @@ test-integration:
 # Removes generated code from tests
 testclean:
 	$(MAKE) -C cmd/_integration-tests clean
+
+.PHONEY testclean test-integration test-go test truss gobindata dependencies
```
hasIan: `PHONY`

hasIan:
```
0.01s$ make test
Makefile:32: *** missing separator.  Stop.
```
:(

Makefile
```diff
@@ -27,3 +28,5 @@ test-integration:
 # Removes generated code from tests
 testclean:
 	$(MAKE) -C cmd/_integration-tests clean
+
+.PHONY testclean test-integration test-go test truss gobindata dependencies
```
hasIan: `.PHONY:`

README.md
```diff
@@ -17,43 +17,36 @@ source.
...
 
 	```
 	go get -u github.com/golang/protobuf/{proto,protoc-gen-go}
 	```
-
-	The compiler plugin, protoc-gen-go, will be installed in `$GOPATH/bin`. 
-It must be in your `$PATH` for the protocol compiler, `protoc`, to find it.
-4. Install google's protobuf/grpc http annotations option. `protoc` will use 
-these to compile http annotations. To install run:
+1. For http annotations install
```
hasIan: Make the language match the rest of the steps.

1. Install...
2. Install...
3. For X install
4. Install...

hasIan: And capitalize HTTP.

Makefile
```diff
@@ -3,8 +3,9 @@
 # Build native Truss by default.
 default: truss
 
-dependencies: 
-	go install github.com/golang/protobuf/protoc-gen-go
+dependencies:
+	go get -u google.golang.org/genproto
+	go get -u github.com/golang/protobuf/{proto,protoc-gen-go}
```
hasAdamr: If these are being added I believe `go-kit/kit` should be as well. As `protoc-gen-go` is the only truss dependency here. The rest `proto` and `genproto` are only dependencies of the services that truss generates.

README.md
```diff
@@ -17,43 +17,36 @@ source.
...
-Otherwise [install from source.](https://github.com/google/protobuf/releases)
-2. Of course, install the Go compiler and tools from https://golang.org/.
-See https://golang.org/doc/install for details.
-3. Install the `protoc-gen-go` and `proto` packages for Go. The simplest 
-way is to run 
+1. Install protoc 3 or newer. The easiest way is to
+download a release from [github](https://github.com/google/protobuf/releases)
+and add to `$PATH`.
+Otherwise [install from source.](https://github.com/google/protobuf)
+1. Install the `protoc-gen-go` and `proto` packages. (Must be in `$PATH`)
```
hasAdamr: Please specify that `protoc-gen-go` needs to be in your path. 

> (The `protoc-gen-go` binary must be in `$PATH`)

Makefile
zaquestion: fixed

README.md
zaquestion: fixed

hasAdamr (commented):
Couple thoughts

hasIan:
While this wasn't introduced in this PR, the use of `go get -u` is sort of bad citizenry. This will potentially modify packages in the user's GOPATH. We can argue that a user should be vendoring their dependencies, but it's still assumptive to potentially update a user's installed packages as part of our build process.

I suggest we use unadorned `go get`, and if a user's dependencies are out of date, they'll have trouble running Truss and we can have a FAQ that suggests updating the dependencies, and the user can determine how to manage this, not our Makefile.

README.md
zaquestion: > (Must be in `$PATH`)
 
Is not sufficient?

hasAdamr: It is a bit confusing because you say to install both `protoc-gen-go` and `proto` but only one of those is actually a binary.

README.md
```diff
@@ -17,43 +17,36 @@ source.
...
-Otherwise [install from source.](https://github.com/google/protobuf/releases)
-2. Of course, install the Go compiler and tools from https://golang.org/.
-See https://golang.org/doc/install for details.
-3. Install the `protoc-gen-go` and `proto` packages for Go. The simplest 
-way is to run 
+1. Install protoc 3 or newer. The easiest way is to
+download a release from [github](https://github.com/google/protobuf/releases)
+and add to `$PATH`.
+Otherwise [install from source.](https://github.com/google/protobuf)
+1. Install the `proto` and `protoc-gen-go` packages. (Later must be in `$PATH`)
```
hasAdamr: Later?

Why not just `protoc-gen-go`?

zaquestion: No particular reason, I just didn't want to repeat the name again and its really just a reminder anyhow.

hasAdamr: If someone is installing go for the first time I am not sure if they would know that one of those is a binary and one of those is just a library.

Repeating the name is fine. I helped people install go and truss last week, the explanation that was there worked perfectly. The current instructions would confuse me as I would wonder where the `proto` binary is at.

hasAdamr:
LGTM \ʕ◔ϖ◔ʔ/

Issue #107, opened 2016-12-03
Unfriendly error messages


hasIan:
While trying to run Truss earlier today, I got this error message. I think I would have understood more, though, if it just crashed.

> cannot parse input definition proto files: cannot create svcdef: error parsing service "HelloServiceClient": Couldn't create service method "SayHello" of service "HelloServiceClient": RequestType creation of service method "SayHello" failed: could not create FieldType, star.Type is not *ast.Ident

😠 

This is a general call to producing more useful error messages for users. Sending a four-level deep error back to the error and vomiting it onto the CLI may in fact do more harm than good. We need to "interpret" the errors that bubble up from lower levels and not just necessarily `errors.Wrap` them, lest we end up merely with a stacktrace as a string with no filenames and line numbers.

hasAdamr:
Could you give steps to reproduce?

hasIan:
Unfortunately, I have no idea what I did to get this error message; I was trying to put together a slapdash presentation and trying to build the simplest example I could, and got this error message. Any recollection of the context was wiped away by the need to get something working in a matter of minutes. I'll try a couple things, but am not sure I'll be able to reproduce it. But reproduction isn't the most important thing here--the point is that the default manner of "handling" errors in most software projects is to pass it up the chain until it reaches the top, which explodes with whatever has been passed up. And as we can see here, it yields something that is no help whatsoever. Our blind adherence to the church of `errors.Wrap` does us a disservice here.

hasIan:
It's not something that I expect a concentrated effort to "fix" this, but as we evolve the project, we should keep this sort of thing in mind.

hasAdamr:
I can quickly add a `-v` or `--verbose` flag so that `github.com/pkg/errors` will print out the stack trace, files, and line numbers that caused the errors. For a developer, that error provides a lot of context.

Adding the line numbers is the difference between `fmt.Println("%v\n", err)` and `fmt.Println("%+v\n", err)`.

hasAdamr:
@hasLeland Please look into adding context that could help the user into the error that Ian got as the root post of this issue. Seems the errors.Wrap was used to add context, but not context that would be helpful to a user to fix their .proto file.

hasIan:
[...]

PR #109, opened 2016-12-06
Add flag for verbose output with strack traces for errors

hasAdamr:
This pull adds a -v flag for verbose mode where debug log messages are printed as well as the full stack trace if an error happens. 

This is in address of issue #107.

### DIFF
--- a/cmd/truss/main.go
+++ b/cmd/truss/main.go
@@ -26,11 +26,10 @@ import (
 var (
 	pbPackageFlag  = flag.String("pbout", "", "The go package path where the protoc-gen-go .pb.go structs will be written.")
 	svcPackageFlag = flag.String("svcout", "", "The go package path where the generated service directory will be written.")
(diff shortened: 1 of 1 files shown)


cmd/truss/main.go
```diff
@@ -311,6 +315,10 @@ func cleanProtofilePath(rawPaths []string) ([]string, error) {
 func exitIfError(err error) {
 	if errors.Cause(err) != nil {
 		defer os.Exit(1)
+		if *verboseFlag {
+			fmt.Printf("%+v\n", err)
```
zaquestion, line 319: Can we remove this check here and just make it the default?

hasAdamr: Have you seen what the stack trace looks like? For most errors it is harder to parse than without the stack trace.

It is only for errors that are cryptic because the errors.Wrap was not used to give useful context.

zaquestion: Seems reasonable.

zaquestion (approved):

PR #110, opened 2016-12-06
Add SHA and build date metadata to Truss binary

hasIan:
See expanded commit message for description.

### DIFF
--- a/cmd/truss/main.go
+++ b/cmd/truss/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
 	"go/build"
@@ -29,10 +30,34 @@ var (
 	svcPackageFlag = flag.String("svcout", "", "The go package path where the generated service directory will be written.")
 )
 
+const (
+	noVersion   string = "<no-version>"
+	noBuildDate string = "<no-build-date>"
+)
+
+var (
+	Version   string = noVersion
+	BuildDate string = noBuildDate
+)
+
 func init() {
 	log.SetLevel(log.InfoLevel)
 
+	buildinfo := new(bytes.Buffer)
+	if Version != "" && Version != noVersion {
+		buildinfo.WriteString("version: ")
+		buildinfo.WriteString(Version)
+		buildinfo.WriteByte(' ')
+	}
+	if BuildDate != "" && BuildDate != noBuildDate {
+		buildinfo.WriteString("built: ")
+		buildinfo.WriteString(strings.Replace(BuildDate, "_", " ", -1))
+	}
+
 	flag.Usage = func() {
+		if buildinfo.Len() > 0 {
+			fmt.Fprintf(os.Stderr, "Truss (%s)\n", strings.TrimSpace(buildinfo.String()))
+		}
 		fmt.Fprintf(os.Stderr, "Usage: %s [OPTION]... [*.proto]...\n", filepath.Base(os.Args[0]))
 		flag.PrintDefaults()
 	}
@@ -41,7 +66,7 @@ func init() {
 
 	if len(flag.Args()) == 0 {
 		fmt.Fprintf(os.Stderr, "%s: missing .proto file(s)\n", filepath.Base(os.Args[0]))
-		fmt.Fprintf(os.Stderr, "Try '%s --help' for more information.\n", filepath.Base(os.Args[0]))
+		flag.Usage()
 		os.Exit(1)
 	}
 }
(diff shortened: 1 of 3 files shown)


cmd/truss/main.go
```diff
@@ -29,8 +30,32 @@ var (
...
+		buildinfo.WriteByte(' ')
+	}
+	if BuildDate != "" && BuildDate != noBuildDate {
+		buildinfo.WriteString("built: ")
+		buildinfo.WriteString(strings.Replace(BuildDate, "_", " ", -1))
+	}
+
 	flag.Usage = func() {
+		if buildinfo.Len() > 0 {
+			fmt.Fprintf(os.Stderr, "Truss (%s)\n", strings.TrimSpace(buildinfo.String()))
```
hasAdamr: I feel it is odd to print out capital `Truss` here when the error message for `flag.Args() == 0` will print out the actual name of the binary, probably `truss`. 

What about using the `filepath.Base(os.Args[0])` here as well?

cmd/truss/main.go
```diff
@@ -39,7 +64,7 @@ func init() {
 
 	if len(flag.Args()) == 0 {
 		fmt.Fprintf(os.Stderr, "%s: missing .proto file(s)\n", filepath.Base(os.Args[0]))
-		fmt.Fprintf(os.Stderr, "Try '%s --help' for more information.\n", filepath.Base(os.Args[0]))
+		flag.Usage()
```
hasAdamr: This was originally done to mirror the output of common command line tools `cp` and `mv` when you call them with no operands.
cp:
```
$ cp
cp: missing file operand
Try 'cp --help' for more information.
```
truss:
```
$ truss
truss: missing .proto file(s)
Try 'truss --help' for more information.
```

Now I feel it is a bit verbose when you don't pass anything
```
$ truss
truss: missing .proto file(s)
Truss (version: 9f6da34c52 built: 2016-12-06 22:49:08 UTC)
Usage: truss [OPTION]... [*.proto]...
  -pbout string
    	The go package path where the protoc-gen-go .pb.go structs will be written.
  -svcout string
    	The go package path where the generated service directory will be written.
  -v	Verbose output with stack traces for errors.
```

What is your reasoning here?

hasAdamr (commented):
Glad to have this info in the binary! It will be nice to be able to see it on someones computer to know if there are on an old truss version.

Some comments about the output with no operands, and how the binary name is printed.

cmd/truss/main.go
hasIan: The user has done something wrong. My thought is, why make them issue another command to figure out what they need to do? I wasn't aware that you had put conscious thought into this to make it match other tools, but it's also notable that a tool like `cp` has many more options. And that the example you cited is GNU specific:

#### BSD on OS X
```
➜  go-truss git:(add_version_tag) ✗ cp
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvX] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-apvX] source_file ... target_directory
```

I'm not going to get too hung up on it one way or the other, though. I like to try to be a steward for the user and not make them do things unnecessarily when I can save them some work, but the unadorned output is not the prettiest, I'll admit.

Since we're talking about passing no options to truss, this isn't related, but I think it'd be nice for Truss to scan the current directory for `*.proto` files and attempt to process them. Since Truss is so directory-oriented as it is, and has opinions on where the file must live (i.e., in the GOPATH), it seems a natural extension.

cmd/truss/main.go
hasIan: In my mind, I was perceiving this as a product banner, where the product is "Truss," which doesn't depend on the name of the binary. But it doesn't matter too much to me. I mean, realistically we're not going to rename `cmd/truss`, so it's all the same, and lowercase is fine. I'll update since I have to fix what I messed up in my "simplify" commit.

hasAdamr:
Once building LGTM.

cmd/truss/main.go
```diff
@@ -24,22 +24,45 @@ import (
...
+	if BuildDate != "" && BuildDate != noBuildDate {
+		buildinfo = fmt.Sprintf("%s built: %s", buildinfo, strings.Replace(BuildDate, "_", " ", -1))
+	}
+
 	flag.Usage = func() {
-		fmt.Fprintf(os.Stderr, "Usage: %s [OPTION]... [*.proto]...\n", filepath.Base(os.Args[0]))
+		if buildinfo != "" {
+			fmt.Fprintf(os.Stderr, "%s (%s)\n", binName, strings.TrimSpace(buildinfo))
+		}
+		fmt.Fprintf(os.Stderr, "Usage: %s [OPTIONS] protofile [protofile2...protofileN]\n", binName)
```
hasIan, line 57: I tried to simplify this a little, but didn't do great. We should make it comply with this http://docopt.org/, which a hybrid of the before and after of this line comes close to.

tunezaq (approved):

PR #117, opened 2016-12-15
Add ability to apply endpoint middlewares granularly

hasIan:
The endpoint middleware strategy forced all endpoint middlewares to be wrapped around every endpoint. This change allows a programmer to selectively apply middlewares to endpoints, as well. See the detailed commit message for more information.

EDIT (TEXT FROM COMMIT):

As it was, InjectEndpointMiddlewares was called for every endpoint on
the defined service. This is fine so long as you want these middlewares
applied to every single endpoint. This breaks down, though, when your
middleware desires are heterogeneous, like, for example, protecting
some endpoints from free access by enforcing some sort of authentication
and authorization, while allowing other endpoints to be freely
accessible (e.g., login, service status, ping).

This change introduces another function into the endpoint middlewares
template, which takes the entire set of endpoints as defined for the
service and returns another set of endpoints. This way, the programmer
has the flexibility wrap different endpoints differently.

This is how service middlewares work, since the middleware wrapping
function takes a service interface, and returns a service interface, but
we should not force a programmer to use Service middlewares if a
middleware would more appropriately be an endpoint middleware.

Additionally, the naming of the middleware application functions has
been changed from using the verb "inject" to the verb "wrap," which
makes more sense in the context of middlewares, which wrap a deeper
functionality.


zaquestion (approved):

gengokit/middlewares/templates/endpoints.go
```diff
@@ -5,9 +5,20 @@ package middlewares
...
 import (
 	"github.com/go-kit/kit/endpoint"
+	svc "{{.ImportPath -}} /generated"
 )
 
-func InjectEndpointMiddlewares(in endpoint.Endpoint) endpoint.Endpoint {
+// WrapEndpoint will be called individually for all endpoints defined in
+// the service. Implement this with the middlewares you want applied to
+// every endpoint.
+func WrapEndpoint(in endpoint.Endpoint) endpoint.Endpoint {
```
hasAdamr: The names of `WrapEndpoint` and `WrapEndpoints` are a bit confusion to me. I thought they were switched until I read the comments. `WrapEndpoints` seems like the natural one that would wrap all of the Endpoint**s**.

Thoughts on `WrapAllEndpoints` and `WrapServiceEndpoints`? 

I know they are more verbose, but seeing "`WrapAllEndpoints`" really gets the message across to me without reading the doc string, but just "`WrapEndpoint`" sounds like it will only wrap one (singular) endpoint.

gengokit/template/NAME-service/NAME-server/server_main.gotemplate
```diff
@@ -71,13 +71,16 @@ func main() {
...
+	// Wrap Endpoints with middlewares. See ../middlewares/endpoints.go
 	endpoints := svc.Endpoints{
 	{{range $i := .Service.Methods -}}
-		{{$i.Name}}Endpoint:    middlewares.InjectEndpointMiddlewares({{ToLower $i.Name}}Endpoint),
+		{{$i.Name}}Endpoint:    middlewares.WrapEndpoint({{ToLower $i.Name}}Endpoint),
 	{{end}}
 	}
 
+	// Wrap selected Endpoints with middlewares. See ../middlewares/endpoints.go
+	endpoints = middlewares.WrapEndpoints(endpoints)
```
hasAdamr, line 80: I love this solution, pass the `svc.Endpoints` struct. I did not think of this at all.

hasAdamr (commented):
Great job Ian, this is a awesome presentation of the problem and your solution.

One comment suggestion some possible name changes. The names of the functions to wrap endpoint middlewares confused me a bit when first reading them. Though the doc string helps enough that it would be fine for this to merge with their current names.

gengokit/middlewares/templates/endpoints.go
hasIan: I'm amenable to more considerate names; I think my main change here was for what I thought was a more metaphorically appropriate verb. The issue I have with `WrapAllEndpoints` here is that it works for both functions, depending on whose perspective you view it through. What `WrapAllEndpoints` or `WrapEndpoint`, as it is now, does, in actuality, is wrap a single endpoint. However, the way it is *used* in the template, it ends up wrapping all the endpoints. But this is naming a function for the way it is used, not what it actually does, and trying to reconcile this makes my head fizzy.

I'm in favor of more clarifying names, but they cannot be subject to interpretation the way the existing ones and the ones in your comment seemingly are (and we're only two people). My mind fails me at the moment, but I will let it marinate, and likewise, make any other proposals you might come up with.

hasAdamr: I have not thought of any great ones yet, but I am going to throw a couple out to get the brain juice flowing.
```
WrapQualifiedEndpoints(...

WrapSelectedEndpoints(...

WrapSpecifiedEndpoints(...

WrapIndividualEndpoints(...
```

From my perspective, these functions should be named in a useful way for the user of truss, the names should describe what the user would get if they filled the function with code. Rather than naming them for what they do. 

Though that is just me, that was the pattern I was going for. 

What about this:
1. Move `func WrapEndpoint(` to `server_main.go`
2. Change the body of `WrapEndpoint` to:
```
func WrapEndpoint(in endpoint.Endpoint) endpoint.Endpoint {
    return middlewares.WrapAllEndpoints(in)
}
```
3. Add `WrapAllEndpoints` to replace `WrapEndpoint`. This way, that function actually does get get called on all Endpoints.

hasLeland: I'll also toss in with a potential name:

- Rename `WrapEndpoint` to `WrapEndpointBlindly`
- Rename `WrapEndpoints` to `WrapEndpointsDiscerning`

I think the addition of a sort of "qualifying adjective" will provide a nice hint into what's going on.

zaquestion: Just throwing in here, upon first review of this PR reading the names of the functions as they are alongside their comments left me with no confusion as to their purpose.

I would even cast a vote to potentially just leave them.

hasIan: None of us has had any great revelations about how this should be named or restructured. I want everyone to be happy with things, but in the name of lot leaving this functional change hanging out to dry, and to keep things moving, I'm going to merge it, and we can deal with it if there is further gnashing of teeth over the way it's been approached. There's certainly a cleaner way that probably only exposes a single function from `endpoints.go`, but it will require thought and discussion that I haven't the luxury of time for right now.

At the end of the day, the user ought not to be a derp, and be capable of reading the GoDoc attached to each function. They will either find the function names intuitive or stupid, and if someone is so moved, they can submit a PR.

hasAdamr: https://github.com/hasIan/go-truss/pull/1

I think I figured out a good middle ground.

PR #118, opened 2016-12-16
Slightly improved error messages

hasLeland:
This is some work to slightly improve error emssages from `svcdef` where possible, and to add one suggestion for the potential source of these errors.

Specifically, `svcdef` will now include file path and line numbers in the error messages it returns, and there is a suggestion that you ensure all protobuf files where passed to truss.


hasAdamr (approved):
🦁

Issue #106, opened 2016-11-30
Proposal: Switch NAME in NAME-service to service name rather than package name


hasAdamr:
Currently if you run truss on:
```
syntax = "proto3";
package Foo;

service Bar {
...
```

You will get a directory structure:
```
./Foo-service/Foo-server/
...etc...
```

It seems silly really, it was just one of the many decisions made during the code restructure of pull #49.

I propose that this is changed to the service name. In this case it would be:
```
./Bar-service/Bar-server/
...etc...
```

This change will cause `./truss/parsepkgname` to be modified to parse the service name rather than the package name. So while this change is pretty simple, it is not trivial.

zaquestion:
:+1: DO IT

hasIan:
This is a good idea. I just ran into this with proto files that have a package with a `pb` suffix, and the service directory ends up being called `myservicepb-service`. Yuck! This can be worked around with conscientious naming of packages, but only if stakeholders agree, so this is a good proposal no matter what.

hasLeland:
Implemented in PR #120 
:+1: :100:

PR #120, opened 2016-12-20
Implement service folder name change proposal #106

hasLeland:
This PR implements the service folder name change proposed in issue #106. From now on, all generated service folders will have a name based on the service name, not the package name.

See issue #106 for more information and examples.


_example/echo.proto
```diff
@@ -10,7 +10,7 @@ package echo;
 
 import "github.com/TuneLab/go-genproto/googleapis/api/serviceconfig/annotations.proto";
 
-service Echo {
+service BounceEcho {
```
hasAdamr: Guessing this was a hand made "test" that got committed in. :p

gendoc/gendoc.go
```diff
@@ -9,6 +9,17 @@ import (
...
 
+func findServiceName(md *deftree.MicroserviceDefinition) string {
+	rv := "default"
+	for _, f := range md.Files {
+		for _, svc := range f.Services {
+			rv = svc.GetName()
+		}
+	}
+	return rv
+
```
hasAdamr: nit: extra whitespace.

gengokit/generator/gen.go
```diff
@@ -31,7 +31,7 @@ func GenerateGokit(sd *svcdef.Svcdef, conf gengokit.Config) (map[string]io.Reade
 	codeGenFiles := make(map[string]io.Reader)
 
 	for _, templPath := range templFiles.AssetNames() {
-		actualPath := templatePathToActual(templPath, sd.PkgName)
+		actualPath := templatePathToActual(templPath, strings.ToLower(sd.Service.Name))
```
hasAdamr: This is a bit gross now, maybe the `ServiceBaseDir` should come in on `conf`. So in  `truss/main.go` conf.ServiceBaseDir` can be set to `strings.ToLower(sd.Service.Name)`, so it seems less magic here.

Was guessing the lowercase path names might cause something odd like this.

Otherwise this deserves a short comment.

truss/parsesvcname/parsesvcname.go
```diff
@@ -0,0 +1,87 @@
+// Package parsesvcname will parse the service name of a protobuf package. The name returned will always be camelcased.
```
hasAdamr: Ah so we are getting the service name from the generated go code from the output of `protoc-gen-go`. So we are getting the camel case.

It is important to note that it will also convert a name from `foo_bar` => `FooBar`. As this will change the folder structure.

```
syntax = "proto3"
service foo_bar {
```
Will create folder `/foobar-service/` rather than `foo_bar-service`.

Not really sure where to note this. 

Thoughts on `log.Warn` when the `ToLower` of the service name in the protofile != `ToLower` of the generated go code service name? This could happen later in the pipeline, but just outputting something so that the user knows that the folder name does not have the underscore.

gendoc/gendoc.go
```diff
@@ -23,7 +34,9 @@ func GenerateDocs(dt deftree.Deftree) map[string]io.Reader {
 	}
 
 	files := make(map[string]io.Reader)
-	files[dt.GetName()+"-service/docs/docs.md"] = strings.NewReader(response)
+	md := dt.(*deftree.MicroserviceDefinition)
+	svcname := strings.ToLower(findServiceName(md))
```
hasAdamr: Does this deal with:

```
syntax = "proto3";
service foo_bar {
```
Will the underscore still be here? As I believe that `package parsesvcname` will not have the `_` because of protoc-gen-go's `CamelCase` function.

truss/parsesvcname/parsesvcname_test.go
```diff
@@ -0,0 +1,114 @@
+package parsesvcname
```
hasAdamr, line 1: Please add a test of the form:
```
syntax = "proto3";
service foo_bar_test {
```

truss/parsesvcname/parsesvcname.go
```diff
@@ -0,0 +1,87 @@
...
+	"strings"
+
+	"github.com/TuneLab/go-truss/svcdef"
+	"github.com/TuneLab/go-truss/truss/execprotoc"
+	"github.com/pkg/errors"
+)
+
+// FromPaths accepts the paths of protobuf definition files and returns the
+// name of the service in that protobuf definition file.
+func FromPaths(gopath []string, protoDefPaths []string) (string, error) {
```
hasAdamr, line 20: So after reading how this works I feel like it might be simpler to just modify `cmd/truss/main.go` to: 
- Start by generating the .pb.go files to either the specified `pbout` flag, or to a temporary directory. 
- With the .pb.go files, create a `svcdef` from the `.pb.go` files. 
    - Either from the temp dir, or from `pbout` flag.
- With the service name
    - Create that service directory
    - Or if a temp dir was used, then create the service directory and move all contents of the temp dir to the service directory.

This way all the files are not generated twice and we also don't create an `svcdef` twice. 

Though since truss is not a long running application, and it is already really fast, maybe this reorganization is an over optimization. And it is fine to do all this "heavy stuff" twice to make the flow from a high level clearer. Though for anything high performance or long running this would not be good, so it strikes me as a bad habit.

@tunezaq , @hasIan , I would like your opinions on if my idea is an over optimization.

hasAdamr (changes requested):
Looks pretty good! I have some worries about service names with underscores `_` in them though.

I also would like to get @tunezaq or @hasIan opinion on the fact that the .pb.go files and `svcdef` are created once and thrown away and then created again. Might be over optimizing but on my own that is rubbing me that bad way.

truss/parsesvcname/parsesvcname_test.go
hasLeland: Test added.

There actually was a case that I hadn't covered that is somewhat related to this that was improperly handled, which was the case when a service begins with an underscore. 

However, I fixed that bug and added a test case for it.

_example/echo.proto
```diff
@@ -10,7 +10,7 @@ package echo;
 
 import "github.com/TuneLab/go-genproto/googleapis/api/serviceconfig/annotations.proto";
 
-service Echo {
+service Bounce_Echo {
```
hasAdamr: Why this name? Seems like an odd name. If it is to show how the folder name will change then it probably should have a comment :).

gengokit/generator/gen.go
```diff
@@ -30,8 +31,12 @@ func GenerateGokit(sd *svcdef.Svcdef, conf gengokit.Config) (map[string]io.Reade
 
 	codeGenFiles := make(map[string]io.Reader)
 
+	// Remove the suffix "-service" since it's added back in by templatePathToActual
+	svcname := strings.TrimSuffix(path.Base(conf.GoPackage), "-service")
```
hasAdamr, line 35: This is better, though actually making a value dedicated to this on `conf` may be nicer. 

It would remove the need for any of these added comments because it would be clear from the code.

Though this is fine too.

hasAdamr (commented):
Thanks for addressing the concerns!

Would still like to see @tunezaq or @hasIan opinion on this pull before merging though.

_example/echo.proto
hasLeland: That was a mistake, which I thought I fixed, that snuck in from debugging. Whoops!

cmd/_integration-tests/cli/cli_test.go
```diff
@@ -218,12 +218,12 @@ func goBuild(name, outputPath, relCodePath string, errChan chan error) {
...
 }
 
-// runServerAndClient execs a TEST-server and TEST-client and puts a
+// runServerAndClient execs a test-server and test-client and puts a
 // runReference to their interaction on the runRefs channel
 func runServerAndClient(path string, port int, debugPort int) runReference {
 	// From within a folder with a truss `service`
 	// These are the paths to the compiled binaries
-	const relativeServerPath = "/bin/TEST-server"
+	const relativeServerPath = "/bin/test-server"
```
hasIan, line 226: Seeing these values called a relative paths, and seeing them start with a slash caused a moment of cognitive dissonance. But this isn't a big deal. :)

hasAdamr (approved):

truss/parsesvcname/parsesvcname.go
hasAdamr: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check: :ballot_box_with_check:

Issue #121, opened 2016-12-20
Truss doesn't expose shutdown hook


hasLeland:
@maxgoz brought up an important point during a chat: truss doesn't expose an "official" way to handle shutdown of a service. Here's a hypothetical situation, based on the brief description @maxgoz gave, as I understand it:

> Within a service, there's a thread pool that takes in units of work and does some kind of operation, such as sending that data to another service. This thread pool needs to be gracefully shut down when the truss service shuts down. Right now, truss doesn't expose a way to hook into it's shutdown routine.

@maxgoz If I got anything incorrect, or you want to expand on anything, please comment!

PR #122, opened 2016-12-21
Add hook for controlling shutdown in truss

hasLeland:
As @maxgoz mentioned in issue #121, there was previously not a way to delay the shutdown of truss, such as for doing cleanup. This PR adds that ability.

## How do I use it?

The signal handler for SIGINT is now within the file `NAME-service/handlers/hooks.go`. You can modify that file however you like and it won't be over-written by `truss`.


hasAdamr (approved):
Oh man, I love it. The diff is so elegant. 

All lines of `-9` are in one template file and the file that generates from it. Amazing.

The elegance is in the simplicity.
:+1: :lion: \ʕ◔ϖ◔ʔ/

gengokit/handler/hooks.go
```diff
@@ -0,0 +1,31 @@
...
+type HookRender struct {
+	prev io.Reader
+}
+
+// Render will return the existing file if it exists, otherwise it will return
+// a brand new copy from the template.
+func (h *HookRender) Render(_ string, _ *gengokit.Data) (io.Reader, error) {
+	if h.prev == nil {
+		return strings.NewReader(hookTempl), nil
+	} else {
```
hasIan, line 28: In Go, if you return inside an if, the preferred style is to omit the `else`, since it's not necessary. Golint/gometalinter will catch stuff like this.

gengokit/handler/templates.go
```diff
@@ -48,3 +48,23 @@ type {{.PackageName}}Service struct{}
...
+package handler
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+)
+
+func InterruptHandler(errc chan<- error) {
```
hasIan, line 61: I feel like the name distracts from the idea that this will handle shutdown. `Shutdown` is a fine name.

Ideally, though, a "fill in the blank" function for the user should be basically empty. The interrupt handler should call an empty function that the the user can populate. The user should not have to maneuver between generated code that's crucial to proper operation, and the blank canvas of adding code. I'd go so far as to say they probably shouldn't even be in the same file. What I had envisioned was something like this.

**hooks.go**
```go
// Shutdown performs any user-defined cleanup tasks that need to be done
// before shutting down the server, like closing connections and any other 
// resource releasing.
func Shutdown() {
}
```
**main.go** (or somewhere else)
```
quitCh := make(chan os.Signal, 1)
signal.Notify(quitCh, syscall.SIGINT, syscall.SIGTERM)
go func() {
    sig := <-quitCh
    hooks.Shutdown()
    errc <- fmt.Errorf("%s", sig)
}()
```
Or even, more logically, I think, since I do not associate shutdown explicitly with signal handling:
```go
func main() {
    defer func() {
        hooks.Shutdown()
        log.Println("exiting")
    }()
    // ...
    go func() {
        errc <- fmt.Errorf("%s", <-quitCh)
    }()
    // ...
    // errc receive allows program execution to flow to the end of the file.
    log.Println(<-errc)
}
// EOF
```

gengokit/template/NAME-service/NAME-server/server_main.gotemplate
```diff
@@ -83,11 +81,7 @@ func main() {
 	ctx := context.Background()
 
 	// Interrupt handler.
-	go func() {
-		c := make(chan os.Signal, 1)
-		signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
-		errc <- fmt.Errorf("%s", <-c)
-	}()
+	go handler.InterruptHandler(errc)
```
hasIan, line 84: `handler.InterruptHandler` is borderline stuttery. It's not `handler.HandleInterrupt`, but the principle is the same: we know it's a handler from the package name. But as mentioned earlier, I'm not especially keen on `InterruptHandler` as the name anyway.

hasIan:
@hasLeland I made some comments post-merge; you guys move more quickly than I can keep on top of things sometimes. Please have a look for consideration in future changes.

gengokit/handler/templates.go
hasLeland: @hasIan I really like the defer shutdown in main pattern you use there. I realize this PR is already merged, but I'll probably create another to implement it that way it's so nice!

PR #128, opened 2017-01-14
Fix #127 - Error on incomplete path parameters

hasLeland:
This is a fix for issue #127 so now an error will be returned when there's an incomplete path, instead of just panicking!

In addition to just the fix, I added unit tests for the `PathParams` function within `gengokit/httptransport`, as well as an integration test for this case.

PR #129, opened 2017-01-14
Build generic `testHTTP` test helper in integration tests

hasLeland:
Last time I added integration tests, @hasIan correctly pointed out that there was some serious copy-pasting going on between certain tests, specifically a function called "testHTTP" which closured around some variables in each parent function to allow for type-agnostic testing. Ian pointed out that each copy-paste of that function added ~20 lines, and there where several tests using it. So much repetition was making the code a bit tough to navigate.

While adding an integration test yesterday I was reminded of that discussion, and while shoveling snow this morning a potential implementation of a more generic version of the same functionality came to me. So I implemented it, and here it is!

### DIFF
--- a/cmd/_integration-tests/transport/http_test.go
+++ b/cmd/_integration-tests/transport/http_test.go
@@ -51,29 +51,11 @@ func TestGetWithQueryRequest(t *testing.T) {
 	var A, B int64
 	A = 12
 	B = 45360
-	want := A + B
-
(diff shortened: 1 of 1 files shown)


hasAdamr:
The new function looks great! But as this is on top of #128 I will fully review it once #128 is in.

adamryman (approved):
📈

hasAdamr (approved):
:1st_place_medal:

PR #130, opened 2017-01-17
Add ability for pbout and svcout flags to be relative directories

hasAdamr:


cmd/truss/main.go
```diff
@@ -94,6 +94,7 @@ func main() {
 // service definition files.
 func parseInput() (*truss.Config, error) {
 	var cfg truss.Config
+	wd, err := os.Getwd()
```
hasIan, line 97: Check `err`.

lelandbatey: While the `err` variable is declared here, if `Getwd` actually did return an error the program would never know since `err` isn't checked and is later overwritten in another function call.

hasLeland, line 97: While the `err` variable is declared here, if `Getwd` actually did return an error the program would never know since `err` isn't checked and is later overwritten in another function call.

hasAdamr: woops

hasLeland:
Seems solid.

## Looks good to me

hasIan (commented):
Looks good, I had this review open a while ago and see it's merged now, and that Leland caught what I had mentioned, so 👍  💯  to you both.

PR #132, opened 2017-01-20
[#131] Supress warnings when httpoptions are not defined

### LINKED ISSUE #131: Reduce warnings when RPCs do not contain httpoptions
Example RPC
```
  rpc Echo(EchoRequest) returns (EchoResp) {}
```

Output
```
$ truss --pbout=. --svcout=../../services *.proto
WARN[0000] Parser found rpc method which lacks HTTP annotations; this is allowed, but will result in HTTP transport not being generated.
WARN[0000] Parser found rpc method which lacks HTTP annotations; this is allowed, but will result in HTTP transport not being generated.
WARN[0000] Error found while parsing file service.proto  error=parser expected 'option' or 'additional_bindings' while parsing options in line '17', instead found '}'
error while parsing http options for the MvEntitiesService service definition
github.com/TuneLab/go-truss/deftree.addHttpOptions
        /home/zaq/projects/go/src/github.com/TuneLab/go-truss/deftree/build_deftree.go:431
github.com/TuneLab/go-truss/deftree.New
        /home/zaq/projects/go/src/github.com/TuneLab/go-truss/deftree/build_deftree.go:86
main.parseServiceDefinition
        /home/zaq/projects/go/src/github.com/TuneLab/go-truss/cmd/truss/main.go:246
main.main
        /home/zaq/projects/go/src/github.com/TuneLab/go-truss/cmd/truss/main.go:79
runtime.main
        /usr/local/go/src/runtime/proc.go:183
runtime.goexit
        /usr/local/go/src/runtime/asm_amd64.s:2086
WARN[0000] Due to the above warning(s), http options and bindings where not parsed and will not be present in the generated documentation.
```

Ideal behavior would no warning when httpoptions are not included and fail if they are defined, but we are missing needed annotations.

zaquestion:
Before if the user did not specify any httpoptions truss would spew a
rather lengthly warning about how http transport would not work. This
change removes those warnings under the assumption if httpoptions are
missing the users probably did so intentionally.

In the case where the annotations needed to parse the httpoptions is not
imported and httpoptions exist. Truss follows existing behavior and dies
immediately.

Resolves #131

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,14 +7,16 @@ SHA := $(shell git rev-parse --short=10 HEAD)
 default: truss
 
 dependencies:
-	go get github.com/go-kit/kit
 	go get github.com/TuneLab/go-genproto
 	go get github.com/golang/protobuf/{proto,protoc-gen-go}
+	go get github.com/jteeuwen/go-bindata/...
+	go get -d github.com/go-kit/kit
 
 update-dependencies:
-	go get -u github.com/go-kit/kit
 	go get -u github.com/TuneLab/go-genproto
 	go get -u github.com/golang/protobuf/{proto,protoc-gen-go}
+	go get -u github.com/jteeuwen/go-bindata/...
+	go get -u github.com/go-kit/kit
 
 # Generate go files containing the all template files in []byte form
 gobindata:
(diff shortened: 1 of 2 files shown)


hasLeland:
Sanity checking indicates that it works great! Seems totally ready to merge!

hasLeland (approved):
Sanity checking indicates that it works great; seems totally ready to merge!

PR #133, opened 2017-01-20
Add flag for static go-bindate modtime to keep template.go from updating

hasAdamr:
With this addition, when `make` is run, the `go-bindata` command will update the modtime to what the flag is set to. Because of this, just running `make` will not make your repo dirty unless you had already edited a template file.

Bonus points if you know what the epoc time means.


gengokit/template/template.go
```diff
@@ -94,7 +94,7 @@ func nameServiceNameCliClientClient_mainGotemplate() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "NAME-service/NAME-cli-client/client_main.gotemplate", size: 2788, mode: os.FileMode(420), modTime: time.Unix(1479253692, 0)}
+	info := bindataFileInfo{name: "NAME-service/NAME-cli-client/client_main.gotemplate", size: 2788, mode: os.FileMode(436), modTime: time.Unix(1464111000, 0)}
```
zaquestion: gonna wanna set the `-mode` as well, suggest `644 (rw-r--r--)`

hasLeland: yeah, mode is the other one that gets me. If that where added, this'd be perfect!

hasAdamr: So `0644 (rw-r--r--)` in decimal:  

```
0*8^3 + 6*8^2 + 4*8^1 + 4*8^0 
=
0*512 + 6*64 + 4*8 + 4*1 
=
0 + 384 + 32 + 4
=
420
```

So go-bindata converts my flag to an int, which is `420` which does mean `(rw-r--r--)` still.

zaquestion (approved):
Very nice!

gengokit/template/gogenerate.go
```diff
@@ -1,4 +1,4 @@
-//go:generate go-bindata -o template.go -pkg template -ignore swp NAME-service/...
+//go:generate go-bindata -mode 0644 -modtime 1464111000 -o template.go -pkg template -ignore swp NAME-service/...
```
hasIan, line 1: 24-May-2016 10:30 AM PDT: Welcome to Tune! :)

PR #134, opened 2017-01-20
Fix #124: Isolate integration test tools locally

### LINKED ISSUE #124: Alter integration tests to not be environment-dependent
The integration tests run `truss` from wherever it is found in the PATH. The tests (including integration tests), should be more tightly constrained to test what is in the working directory and not be affected much by the environment in which it's running.

## Backstory
In my case, I use GVM to manage my Go versions, which sets up separate `GOPATH`s for each Go version (which I think is dumb, at least for my use case--I'm not using it like a Python virtualenv or anything). As such, I alter my GOPATH to point back to my original, pre-GVM one (`$HOME/go/bin`), but GVM also changes the PATH to put its version-specific `$GOPATH/bin` directory at the head of the path. This meant that running `go install` in the Truss directory was placing binaries into `$HOME/bin`, but the integration tests, when relying on the shell to resolve binaries by searching the PATH, were finding--in this case--truss binaries in `$HOME/.gvm/pkgsets/go1.7.4/global/bin`. So the integration tests were running an outdated binary.

## Proposal
Build truss binaries in the integration-test directory and run it explicitly. However, during truss' run, it in turn invokes the protocast binary, which apparently it finds through the PATH. So some more thought is required here.

## In Conclusion
This is not something that should happen often, but it is a slight vulnerability and I just got bit by it. Aiming to isolate our build and test phases from any environmental inconsistencies will serve us well.

hasLeland:
This PR is a basic implementation of the behavior described in issue #124; instead of running integration tests by doing a `go install` of the tools in the repo and *hoping* those tools will then be available in the `$PATH`, we instead install the tools to the a set directory and append that directory to the front of our `$PATH` before running the tests.

This is a somewhat hackish way to ensure the correct `truss` and `protoc-gen-truss-protocast` binaries are used during integration tests, but it works reliably and I couldn't think of another way to do this without a total overhaul ; )

### DIFF
--- a/cmd/_integration-tests/Makefile
+++ b/cmd/_integration-tests/Makefile
@@ -13,19 +13,23 @@ all: test
 
 test: clean test-transport test-cli
 
-local-tools:
-	go build -o protoc-gen-truss-protocast github.com/TuneLab/go-truss/cmd/protoc-gen-truss-protocast
+truss:
 	go build -o truss github.com/TuneLab/go-truss/cmd/truss
 
-test-transport: local-tools
+protoc-gen-truss-protocast:
+	go build -o protoc-gen-truss-protocast github.com/TuneLab/go-truss/cmd/protoc-gen-truss-protocast
+
+test-transport: protoc-gen-truss-protocast truss
 	@which truss
 	@printf '$(TRANSPORT_TEST_MSG)'
 	$(MAKE) -C transport
+	rm -f ./truss ./protoc-gen-truss-protocast
 
-test-cli: local-tools
+test-cli: protoc-gen-truss-protocast truss
 	@which truss
 	@printf '$(CLI_TEST_MSG)'
 	go test -v ./cli
+	rm -f ./truss ./protoc-gen-truss-protocast
 
 clean:
 	rm -f ./truss ./protoc-gen-truss-protocast
--- a/cmd/_integration-tests/Makefile
+++ b/cmd/_integration-tests/Makefile
@@ -7,20 +7,28 @@ TRANSPORT_TEST_MSG=\n$(OK_COLOR)Starting transport end to end test:$(END_COLOR_L
 
 CLI_TEST_MSG=\n$(OK_COLOR)Start server and cliclient generate, build, and run test:$(END_COLOR_LINE)
 
+export PATH := $(CURDIR):$(PATH)
 
 all: test
 
 test: clean test-transport test-cli
 
-test-transport:
+local-tools:
+	go build -o protoc-gen-truss-protocast github.com/TuneLab/go-truss/cmd/protoc-gen-truss-protocast
+	go build -o truss github.com/TuneLab/go-truss/cmd/truss
+
+test-transport: local-tools
+	@which truss
 	@printf '$(TRANSPORT_TEST_MSG)'
 	$(MAKE) -C transport
 
-test-cli:
+test-cli: local-tools
+	@which truss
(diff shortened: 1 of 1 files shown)


cmd/_integration-tests/Makefile
```diff
@@ -7,20 +7,28 @@ TRANSPORT_TEST_MSG=\n$(OK_COLOR)Starting transport end to end test:$(END_COLOR_L
 
 CLI_TEST_MSG=\n$(OK_COLOR)Start server and cliclient generate, build, and run test:$(END_COLOR_LINE)
 
+export PATH := $(CURDIR):$(PATH)
 
 all: test
 
 test: clean test-transport test-cli
```
hasAdamr, line 14: This should probably have a clean at the end of it as well, otherwise these binaries are going to stick around even after the tests pass. Then there are hanging out in the repo untracked untell you call `make testclean`. If the tests pass, the binaries should be removed.

The tests in `cli` and `transport` both clean up at the end if the pass. Though we need something here to remove the binaries.

cmd/_integration-tests/Makefile
```diff
@@ -7,20 +7,28 @@ TRANSPORT_TEST_MSG=\n$(OK_COLOR)Starting transport end to end test:$(END_COLOR_L
 
 CLI_TEST_MSG=\n$(OK_COLOR)Start server and cliclient generate, build, and run test:$(END_COLOR_LINE)
 
+export PATH := $(CURDIR):$(PATH)
```
hasAdamr, line 10: Note that `protoc-gen-truss-protocast` needs to be installed for the normal `go test ./...` tests to work, as they shell out to `protoc` and `protoc-gen-truss-protocast`. So this will only effect the integration tests in this Makefile, and if the user has never installed `protoc-gen-truss-protocast` then then those tests will fail. 

Though that binary should never change, so I am going to say it is a non issue. As this change is mostly so that the developer running the tests run the intergration tests with the current source code on disk rather than whatever truss they have in their own path.

cmd/_integration-tests/Makefile
```diff
@@ -7,20 +7,28 @@ TRANSPORT_TEST_MSG=\n$(OK_COLOR)Starting transport end to end test:$(END_COLOR_L
...
 all: test
 
 test: clean test-transport test-cli
 
-test-transport:
+local-tools:
+	go build -o protoc-gen-truss-protocast github.com/TuneLab/go-truss/cmd/protoc-gen-truss-protocast
+	go build -o truss github.com/TuneLab/go-truss/cmd/truss
+
+test-transport: local-tools
```
hasAdamr: Does having `local-tools` before `test-transport` and `test-cli` mean that these get built twice? Or is `make` smart enough to know?

hasAdamr (commented):
Looking good! I'm sure it will help somebody not have headaches in the future. @hasian :face_with_head_bandage: 

Couple comments, should be pretty easy.

cmd/_integration-tests/Makefile
hasLeland: Make is only able to figure it out if a rule generates a file in the current directory of the same name, which this doesn't, so yes, it will be built twice. Fixing that.

hasAdamr (approved):

cmd/_integration-tests/Makefile
hasLeland: The clean for these has been added!

PR #136, opened 2017-01-21
Move to CSV-style query params for basic repeated types

hasLeland:
This PR changes the behavior of generated clients and servers such that query parameters of basic repeated types will no longer require surrounding brackets. This is done for greater compatibility, as most applications expect the following query string format:

    /some/path?foo=bar,baz

As opposed to the previous behavior:

    /some/path?foo=[bar,baz]

This is also done so that newly-generated clients and servers may be compatible with existing infrastructure, in our case, the `multiverse` project.

### DIFF
--- a/cmd/_integration-tests/transport/http_test.go
+++ b/cmd/_integration-tests/transport/http_test.go
@@ -89,7 +89,7 @@ func TestGetWithRepeatedQueryRequest(t *testing.T) {
 
 	testHTTP(t, &resp, &expects, nil, "GET", "getwithrepeatedquery?%s=[%d,%d]", "A", A[0], A[1])
 	// csv style
-	//testHTTP(nil, "GET", "getwithrepeatedquery?%s=%d,%d", "A", A[0], A[1])
+	testHTTP(t, &resp, &expects, nil, "GET", "getwithrepeatedquery?%s=%d,%d", "A", A[0], A[1])
 	// multi / golang style
 	//testHTTP(nil, "GET", "getwithrepeatedquery?%s=%d&%s=%d]", "A", A[0], "A", A[1])
 }
--- a/gengokit/httptransport/httptransport.go
+++ b/gengokit/httptransport/httptransport.go
@@ -260,9 +260,19 @@ err = json.Unmarshal([]byte({{.LocalName}}Str), {{.LocalName}})`
 		// All repeated args of any type are represented as slices, and bare
 		// assignments to a slice accept a slice as the rvalue. As a result,
 		// LocalName will be declared as a slice, and json.Unmarshal handles
-		// everything else for us.
+		// everything else for us. Addititionally, if a type is a Base type and
+		// is repeated, we first attempt to unmarshal the string we're
+		// provided, and if that fails, we try to unmarshal the string
+		// surrounded by square brackets. If THAT fails, then the string does
+		// not represent a valid JSON string and an error is returned.
 		repeatedUnmarshalTmpl := `
 var {{.LocalName}} {{.GoType}}
+{{- if and (and .IsBaseType .Repeated) (not (Contains .GoType "[]byte"))}}
+err = json.Unmarshal([]byte({{.LocalName}}Str), &{{.LocalName}})
+if err != nil {
+	{{.LocalName}}Str = "[" + {{.LocalName}}Str + "]"
+}
+{{- end}}
 err = json.Unmarshal([]byte({{.LocalName}}Str), &{{.LocalName}})`
 		errorCheckingTmpl := `
 if err != nil {
@@ -276,7 +286,7 @@ if err != nil {
 			preamble = repeatedUnmarshalTmpl
 		}
 		jsonConvTmpl := preamble + errorCheckingTmpl
-		code, err := ApplyTemplate("UnmarshalNonBaseType", jsonConvTmpl, f, nil)
+		code, err := ApplyTemplate("UnmarshalNonBaseType", jsonConvTmpl, f, TemplateFuncs)
 		if err != nil {
 			panic(fmt.Sprintf("Couldn't apply template: %v", err))
 		}
@@ -365,9 +375,10 @@ func LowCamelName(s string) string {
 // TemplateFuncs contains a series of utility functions to be passed into
 // templates and used within those templates.
 var TemplateFuncs = template.FuncMap{
-	"ToLower": strings.ToLower,
-	"Title":   strings.Title,
-	"GoName":  gogen.CamelCase,
+	"ToLower":  strings.ToLower,
+	"Title":    strings.Title,
+	"GoName":   gogen.CamelCase,
+	"Contains": strings.Contains,
 }
 
 // ApplyTemplate applies a template with a given name, executor context, and
--- a/gengokit/httptransport/templates.go
+++ b/gengokit/httptransport/templates.go
@@ -46,6 +46,8 @@ var ClientEncodeTemplate = `
 	// the http request (path, query, and body).
 	func EncodeHTTP{{$binding.Label}}Request(_ context.Context, r *http.Request, request interface{}) error {
 		fmt.Printf("Encoding request %v\n", request)
+		strval := ""
+		_ = strval
 		req := request.(*pb.{{GoName $binding.Parent.RequestType}})
 		_ = req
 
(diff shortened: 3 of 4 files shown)


gengokit/httptransport/templates.go
```diff
@@ -46,6 +46,8 @@ var ClientEncodeTemplate = `
 	// the http request (path, query, and body).
 	func EncodeHTTP{{$binding.Label}}Request(_ context.Context, r *http.Request, request interface{}) error {
 		fmt.Printf("Encoding request %v\n", request)
+		strval := ""
+		_ = strval
```
zaquestion, line 50: ?

gengokit/httptransport/templates_test.go
```diff
@@ -62,6 +62,8 @@ func TestGenClientEncode(t *testing.T) {
 // the http request (path, query, and body).
 func EncodeHTTPSumZeroRequest(_ context.Context, r *http.Request, request interface{}) error {
 	fmt.Printf("Encoding request %v\n", request)
+	strval := ""
+	_ = strval
```
zaquestion, line 66: ?

gengokit/httptransport/templates_test.go
```diff
@@ -62,6 +62,8 @@ func TestGenClientEncode(t *testing.T) {
...
 	req := request.(*pb.SumRequest)
 	_ = req
```
zaquestion, line 68: Consider `_ = request.(*pb.SumRequest)`

gengokit/httptransport/httptransport.go
```diff
@@ -260,9 +260,19 @@ err = json.Unmarshal([]byte({{.LocalName}}Str), {{.LocalName}})`
...
-		// everything else for us.
+		// everything else for us. Addititionally, if a type is a Base type and
+		// is repeated, we first attempt to unmarshal the string we're
+		// provided, and if that fails, we try to unmarshal the string
+		// surrounded by square brackets. If THAT fails, then the string does
+		// not represent a valid JSON string and an error is returned.
 		repeatedUnmarshalTmpl := `
 var {{.LocalName}} {{.GoType}}
+{{- if and (and .IsBaseType .Repeated) (not (Contains .GoType "[]byte"))}}
+err = json.Unmarshal([]byte({{.LocalName}}Str), &{{.LocalName}})
```
zaquestion, line 271: This feels heavy.. Granted the happy path remains unaffected this way, is there another reason for not doing something more like?
```
	if !(s[0] == '[' || s[len(s)-1] == ']') {
		s = "[" + s + "]"
	}
```
https://play.golang.org/p/TwKz5zXVI7

cmd/_integration-tests/transport/http_test.go
```diff
@@ -87,9 +87,9 @@ func TestGetWithRepeatedQueryRequest(t *testing.T) {
 		V: A[0] + A[1],
 	}
 
-	testHTTP(t, &resp, &expects, nil, "GET", "getwithrepeatedquery?%s=[%d,%d]", "A", A[0], A[1])
+	//testHTTP(t, &resp, &expects, nil, "GET", "getwithrepeatedquery?%s=[%d,%d]", "A", A[0], A[1])
```
zaquestion: We should still test this one, yeah?

gengokit/httptransport/templates.go
hasAdamr: The `_ = strval` is here for cases where there are no path parameters to be parsed, if there are no path parameters to be parsed then just having `strva := ""` would cause a compiler error as it is unused.

gengokit/httptransport/templates_test.go
hasAdamr: Depending on what http annotations are used, `req` will be used or unused. Same reason as https://github.com/TuneLab/go-truss/pull/136#discussion_r97655106

gengokit/httptransport/httptransport.go
hasAdamr: Ah, don't try the `json.Unmarshal` twice, just check of the `'['` and add them if they are not there then do the `json.Unmarshal`.

hasLeland: The reason that json unmarshalling happens twice is because for certain values of certain types, they're automatically marshelled to string representations which aren't and should not be surrounded by square brackets. For example:

Let's say you're making a request, and one field on that request is a repeated series of ints. If you don't include anything in that field, then when it is marshalled into a string, it won't result in a string of `[]`, it will instead be marshalled into a string of `null`.

That's why I chose to implement this by attempting decoding twice, instead of via blind string manipulation. If you guys think there's a better way to do it, I do want to know.

gengokit/httptransport/templates_test.go
hasLeland: These variables are declared this way because they're variables that need to exist, but depending on the types of the variables we decode in this method, code may not be generated which ever actually uses these variables, which would cause the compiler to fail to compile this generated code. So instead, we must declare them and then tell the compiler to not complain if that variable is ever unused by assigning it to underscore.

gengokit/httptransport/httptransport.go
zaquestion: @hasLeland Is there actually another case outside of the null case? Presumably any repeated type is going to be `null` or `[1,..,n]`, yeah?
```
		if len(v) > 0 && !(v[0] == '[' || v[len(v)-1] == ']') && v != "null" {
			v = "[" + v + "]"
		}
```
Definitely makes the resulting if a tad gnarlier, _probably_ still cheaper than multiple decodes. Particularly if we suspect `csv` style to be a common case, failing through for each one can really add up. I'm curious what @hasIan has to say though. 

https://play.golang.org/p/y9CNqU18-D

hasAdamr: This is probably still cheaper than reflecting twice.

As this happens per request it would be good to be efficient here, especially if someone is sending a lot of json requests without the `[]`'s.

Though a switch in the order in your condition @zaquestion will add even more speed!

```
// Short circuit if v != "null"
if len(v) > 0 && v != "null" && !(v[0] == '[' || v[len(v)-1] == ']') {
```

hasIan: Okay guys, I got way too involved here and decided to try some science. Well, pseudoscience. I made up some benchmarks, with bogus input data, to try to suss out the best of a few approaches.

Check out the results here: https://github.com/ianfoo/go-json-unmarshal-experiment

Long story short, it looks like using a heuristic to determine whether the input needs to be wrapped with square brackets is the fastest and requires the least allocations. There may possibly be a better approach, but out of the ones we've talked about, checking and adjusting, then unmarshaling, wins.

This said, the boolean check in this thread may not be what we want. Check the results here: https://play.golang.org/p/tnB3U_c-ua. The boolean for checking square brackets here will indicate that an input does not need to be wrapped if *either* begins or ends with a square bracket (or both). The one that I have used will only indicate that an input needs to be wrapped if it begins *and* ends with square brackets.

hasLeland: @hasIan Well, I have never had anything so thoroughly answered, thank you Ian! Going to implement that now.

hasAdamr: @zaquestion @hasIan @hasLeland, 

I now am back to thinking that Lelands original method of unmarshaling twice is the correct response. 

This is because the truss generated client code WILL ALWAYS send json on http transport. Because of this, just unmarshaling it without a check will always work with truss clients. 

We should be optimizing for truss to truss communication. While it would be nice to support this without the `[]`s, we should expect them as truss clients will send them. Therefore just unmashal and then try something else.

Note: This whole pull request was because I wanted to have new truss servers communicate with Lee's truss clients. I have sense realized that this is even less possible, as even with this change every string still requires quotes around it, and Lee's services do not send them with quotes.

So could we merge this as is? Less work for Leland and it will be more efficent in the truss to truss case which should be at least the 90% plus case.

hasAdamr (approved):

gengokit/httptransport/httptransport.go
hasLeland: Currently in this PR, truss-to-truss actually will *not* send JSON, it will strip off the square brackets before sending. I will remove the portion which strips braces before sending, as that will keep things nice and fast, like you say.

hasAdamr: @hasLeland, wait really? How was this working before then? Or did that change in this pull?

hasLeland: @hasAdamr That changed in this pull. See here: https://github.com/TuneLab/go-truss/pull/136/files#diff-0fe7c6aec1f529107babf5f63bddb542R79

hasLeland: @hasAdamr Alright, that change has been pushed and passes the tests. So this PR will work the way you explained above.

hasAdamr (approved):

PR #138, opened 2017-01-21
Greatly reduce docker image size and build time

zaquestion:
This addresses the long standing PR \#69 and frustrations around obtaining a
more usable docker image. Note: On linux output files are still returned as
root as of this PR. I'm on the side of bringing in these improvements and
opening a "low hanging fruit" issue that someone can grab.


hasAdamr (approved):
Too bad about this outputting root files on linux still. 

Though this is a huge improvement on what we have currently.

I am happy to see that Alpine now has a more recent version of `protoc`. When I originally was trying to build this Alpine only had `protoc 2.9.1` and so we had to build from source.

**NOTE**: When you merge please add the:
>  "low hanging fruit" issue that someone can grab.

PR #139, opened 2017-01-24
Fix relative directory flags for --pbout and --svcout

hasAdamr:
It seems that relative directories were being put as imports with truss if you used a relative directory as in the `--svcout` or `--pbout` flag. 

So `--svcout ../service` would add imports of the form `../service/truss-service/generated`... Which can cause all kinds of issues.

I have fixed that and have added to a test for reach flag for a relative directory to make sure the that the go tools can always deal with the output of a relative directory flag.


hasLeland:
## Looks solid :+1:

PR #141, opened 2017-01-26
Fix incorrect naming of body parameters in client encoding

hasLeland:
Within the client encoder, body parameters were being encoded incorrectly using the camelCased version of their go-name, instead of their original protobuf name.

This PR modifies the encoding of client parameters to use the built in protobuf structs encoded to json instead of a map of strings encoded to json, as the structs contain the correct struct tags to do this automatically. Additionally, it adds tests for catching this issue.

Big thanks to @siim- for pointing out this bug!

### DIFF
--- a/cmd/_integration-tests/transport/handlers/server/server_handler.go
+++ b/cmd/_integration-tests/transport/handlers/server/server_handler.go
@@ -83,7 +83,7 @@ func (s transportService) GetWithPathParams(ctx context.Context, in *pb.GetWithQ
 	return &response, nil
 }
 
-// EchoOddNames implements Service. Why do we always put this comment here?
+// EchoOddNames implements Service.
 func (s transportService) EchoOddNames(ctx context.Context, in *pb.OddFieldNames) (*pb.OddFieldNames, error) {
 	return in, nil
 }
--- a/cmd/_integration-tests/transport/handlers/server/server_handler.go
+++ b/cmd/_integration-tests/transport/handlers/server/server_handler.go
@@ -83,6 +83,11 @@ func (s transportService) GetWithPathParams(ctx context.Context, in *pb.GetWithQ
 	return &response, nil
 }
 
+// EchoOddNames implements Service. Why do we always put this comment here?
+func (s transportService) EchoOddNames(ctx context.Context, in *pb.OddFieldNames) (*pb.OddFieldNames, error) {
+	return in, nil
+}
+
 var testError error = errors.New("This error should be json over http transport")
 
 // ErrorRPC implements Service.
--- a/cmd/_integration-tests/transport/http_test.go
+++ b/cmd/_integration-tests/transport/http_test.go
@@ -158,6 +158,26 @@ func TestCtxToCtxViaHTTPHeaderClient(t *testing.T) {
 	}
 }
 
+func TestEchoOddNamesClient(t *testing.T) {
+	var req pb.OddFieldNames
+	req.CamelCase = 12
+	req.SnakeCase = 24
+	req.XWhy_So_Many_Underscores = 36
+
+	svchttp, err := httpclient.New(httpAddr)
+	if err != nil {
+		t.Fatalf("failed to create httpclient: %q", err)
+	}
+
+	resp, err := svchttp.EchoOddNames(context.Background(), &req)
+	if err != nil {
+		t.Fatalf("httpclient returned error: %q", err)
+	}
+	if !reflect.DeepEqual(resp, &req) {
+		t.Fatalf("Expected req and resp to be identical, instead: \n%+v\n%+v", req, *resp)
+	}
+}
+
 func TestCtxToCtxViaHTTPHeaderRequest(t *testing.T) {
 	var resp pb.MetaResponse
 	var key, value = "Truss-Auth-Header", "SECRET"
--- a/cmd/_integration-tests/transport/http_test.go
+++ b/cmd/_integration-tests/transport/http_test.go
@@ -158,11 +158,14 @@ func TestCtxToCtxViaHTTPHeaderClient(t *testing.T) {
 	}
 }
 
+// Test different kinds of message field names from protobuf definition.
+// e.g. "CamelCase", "snake_case", "__why_so_many_underscores"
 func TestEchoOddNamesClient(t *testing.T) {
-	var req pb.OddFieldNames
-	req.CamelCase = 12
-	req.SnakeCase = 24
-	req.XWhy_So_Many_Underscores = 36
+	req := pb.OddFieldNames{
+		CamelCase:                12,
+		SnakeCase:                24,
+		XWhy_So_Many_Underscores: 36,
+	}
 
 	svchttp, err := httpclient.New(httpAddr)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


gengokit/template/template.go
```diff
@@ -279,7 +279,7 @@ func nameServiceHandlersServerServer_handlerGotemplate() (*asset, error) {
 	return a, nil
 }
 
-var _nameServiceMiddlewaresEndpointsGotemplate = []byte("\x1f\x8b\x08\x00\x00\x09\x6e\x88\x00\xff\x04\xc0\x41\x0a\xc4\x20\x0c\x05\xd0\xfd\x9c\x22\x6b\x61\x26\x87\x99\x13\x48\xf3\x1b\xa4\x6a\xc4\xa4\x74\x21\xde\xbd\x6f\x2d\x4e\xf4\x07\x48\xed\x1b\xf3\x76\x67\x45\x57\xbb\x4a\x70\x2b\x22\x15\x4f\x9e\x70\x0e\xb4\x51\x73\xc0\x19\x5d\x86\x95\x1e\xfe\x53\xa3\xd3\x26\x1d\x26\xa0\xc4\x7b\x7f\xde\x00\x00\x00\xff\xff\x80\x45\x66\x4f\x4c\x00\x00\x00")
+var _nameServiceMiddlewaresEndpointsGotemplate = []byte("\x1f\x8b\x08\x00\x00\x09\x6e\x88\x00\xff\x0c\xc6\xc1\x0d\xc3\x20\x0c\x05\xd0\x7b\xa7\xf0\x19\xa9\xf5\x30\x9d\x00\xd5\xbf\x16\x0a\x60\x84\x1d\xe5\x80\xd8\x3d\xb9\xbd\xb5\x38\xd1\x17\x20\xb5\x77\xcc\xd3\x9d\x15\x5d\xed\x28\xc1\xad\x88\x54\x5c\x79\xc2\x39\xd0\x46\xcd\xf1\x08\x5d\x86\x95\x1e\xfe\x51\xa3\xbf\x4d\xfa\x99\x80\x12\xef\xfd\xba\x03\x00\x00\xff\xff\x80\x45\x66\x4f\x4c\x00\x00\x00")
```
hasAdamr: What changed here? I don't see `NAME-service/middlewares/endpoints.go` committed.

cmd/_integration-tests/transport/http_test.go
```diff
@@ -158,6 +158,26 @@ func TestCtxToCtxViaHTTPHeaderClient(t *testing.T) {
 	}
 }
 
+func TestEchoOddNamesClient(t *testing.T) {
```
hasAdamr: We should get in the habit of putting a small comment on why a test was added. Just so we can see in the future while scrolling through.

Please either make a github issue about this and reference it or add a small comment like:

```
// Test different kinds of message field names from protobuf definition.
// e.g. "CamelCase", "snake_case", "__why_so_many_underscores"
```

hasAdamr (commented):
Looks awesome!

Fairly simple too :+1: 

I want to know what changed in the `template.go` file before I hit the approve, but as soon as I know why I will hit approve :).

cmd/_integration-tests/transport/handlers/server/server_handler.go
```diff
@@ -83,6 +83,11 @@ func (s transportService) GetWithPathParams(ctx context.Context, in *pb.GetWithQ
 	return &response, nil
 }
 
+// EchoOddNames implements Service. Why do we always put this comment here?
```
hasIan: We want to generate code that passes `golint` cleanly. The comment could be made more useful, but exported functions that do not have a documentation comment that begins with the function name will be flagged by golint.

gengokit/template/template.go
hasLeland: I suspect this is the result of an outdated `go-bindata`. Updating, re-running and pushing now to confirm.

cmd/_integration-tests/transport/http_test.go
```diff
@@ -158,6 +158,28 @@ func TestCtxToCtxViaHTTPHeaderClient(t *testing.T) {
 	}
 }
 
+// Test different kinds of message field names from protobuf definition.
+// e.g. "CamelCase", "snake_case", "__why_so_many_underscores"
+func TestEchoOddNamesClient(t *testing.T) {
+	var req pb.OddFieldNames
```
hasIan: This is just a stylistic thing, and not something I care if you change, but more typically, I think, are single-statement initializations, versus multiple-statement initializations like so:
```go
req := pb.OddFieldNames{
    CamelCase: 12,
    SnakeCase: 24,
    XWhy_So_Many_Underscores: 36,
}
```

gengokit/httptransport/templates.go
```diff
@@ -84,13 +84,14 @@ var ClientEncodeTemplate = `
 
 		// Set the body parameters
 		var buf bytes.Buffer
-		toRet := map[string]interface{}{
+		toRet := request.(*pb.{{GoName $binding.Parent.RequestType}})
 		{{- range $field := $binding.Fields -}}
 			{{if eq $field.Location "body"}}
-				"{{$field.CamelName}}" : req.{{$field.CamelName}},
+				{{/* Only set the fields which should be the body, so all
+				others will be omitted due to emptyness */}}
```
hasIan: *emptiness

gengokit/template/template.go
hasLeland: Yep, it was an outdated version of `go-bindata`. The update followed by re-running causes the binary data of those files to fall back in line with `master`.

gengokit/httptransport/templates.go
```diff
@@ -84,13 +84,14 @@ var ClientEncodeTemplate = `
 
 		// Set the body parameters
 		var buf bytes.Buffer
-		toRet := map[string]interface{}{
+		toRet := request.(*pb.{{GoName $binding.Parent.RequestType}})
 		{{- range $field := $binding.Fields -}}
 			{{if eq $field.Location "body"}}
-				"{{$field.CamelName}}" : req.{{$field.CamelName}},
+				{{/* Only set the fields which should be the body, so all
```
hasIan: This comment confuses me.

hasLeland: What's going on is that we are making a request to the server, and that request contains fields. However, an endpoint may only want *certain* fields in the body of the POST to that server.

At this point in code generation, we have the struct form of the data which should be sent to the server. What I'm trying to convey is that by creating a new struct and only copying the fields which should be sent as part of the body into it, we can marshal the new struct to string form and get a json message which only contains the fields we copied, which are the fields we want to send in the body.

The potential problem with this approach is that the unset fields still contain their default value, which may be encoded and sent in the body. However, since all generated struct fields have struct tags indicating `omitempty`, struct fields with uninitialized values won't show up in the json. Thus, the one potential problem with the approch is avoided.

---------

That's pretty long though, so I used the short comment that you see there, as terse as it is. Do you have any recommendations for a better comment?

cmd/_integration-tests/transport/handlers/server/server_handler.go
hasLeland: Alright, makes sense. I'll remove that little tidbit.

hasAdamr (approved):

PR #144, opened 2017-01-31
Add help flag with description, add spacing for help

### LINKED ISSUE #34: The '--help' flag for truss should give directions
Currently, the output of truss is quite bare, and doesn't include at all how you'd run this on a set of proto files, nor does it include what truss is, how to find more information. Currently it looks like this:

```
Usage of truss:
  -nobuild
        Set -nobuild to generate code without building
```

And it should include things like:
- How to pass in proto files
- What `truss` is meant to do
- Where to find more about `truss` (a link back to this github, for example)

hasAdamr:
Fixes #34.

In that issue @hasLeland recommened putting a way for the user to learn more, though most tools do not do this as developers are good at searching for thing. 

Docker though has a nice output:
> $ docker
> A self-sufficient runtime for containers

The output for `truss -h` looks like this:
```
$ truss -h
truss (version: c559a9d19f built: 2017-01-31 22:38:27 UTC)

Usage: truss [options] <protofile>...

Generates go-kit services using proto3 and gRPC definitions.

Options:
  -h, --help            Print usage
      --pbout string    Go package path where the protoc-gen-go .pb.go files will be written
      --svcout string   Go package path where the generated Go service will be written
  -v, --verbose         Verbose output
```

### DIFF
--- a/cmd/truss/main.go
+++ b/cmd/truss/main.go
@@ -26,7 +26,8 @@ import (
 var (
 	pbPackageFlag  = flag.String("pbout", "", "Go package path where the protoc-gen-go .pb.go files will be written")
 	svcPackageFlag = flag.String("svcout", "", "Go package path where the generated Go service will be written")
-	verboseFlag    = flag.BoolP("verbose", "v", false, "verbose output")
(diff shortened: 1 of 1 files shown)


zaquestion (approved):
A++++++++

PR #150, opened 2017-02-04
Update env vars to always beat flags

hasAdamr:


gengokit/template/NAME-service/NAME-server/server_main.gotemplate
```diff
@@ -28,21 +28,23 @@ func main() {
 		httpAddr       = flag.String("http.addr", ":5050", "HTTP listen address")
 		grpcAddr       = flag.String("grpc.addr", ":5040", "gRPC (HTTP) listen address")
 	)
-	flag.Parse()
 
 	// Use environment variables, if set. Flags have priority over Env vars.
-	if os.Getenv("DEBUG_ADDR") != ""  && *debugAddr == ":5060" {
+	if os.Getenv("DEBUG_ADDR") != "" {
```
hasIan: What about
```go
if addr := os.Getenv("DEBUG_ADDR"); addr != "" {
    *debugAddr = addr
}
```
(Repeat as necessary for other settings.)

hasIan (approved):
Minor cleanup suggested. Your description is backward though, if I'm reading "beat" the way I think you mean it. As implemented, flags will supersede environment variables, as it should be.

hasLeland:
Very clean; and excellent revision.

Looks good to me.

PR #151, opened 2017-02-14
Implement flag-sets for generated CLI clients

hasLeland:
This PR changes the way generated CLI clients parse command line flags. The previous output was rather verbose:

```
$ foo -help
Usage of foo-cli-client:
  -bar.alpha int
  -bar.beta int
  -bar.epsilon int
  -zip.delta string
  -zip.alpha string
  -zip.theta string
  -grpc.addr string
    	gRPC (HTTP) address of addsvc (default ":5040")
  -http.addr string
    	HTTP address of addsvc
  -method string
        bar,zip
```

The consequence of these long command line names was very redundent invocations e.g.:

    foo -method bar -bar.alpha 1 -bar.beta 2 -bar.epsilon 3

---------

This PR changes the command line interface to instead be:

```
$ foo -help
Usage of foo-cli-client:
  -grpc.addr string
    	gRPC (HTTP) address of addsvc (default ":5040")
  -http.addr string
    	HTTP address of addsvc
Subcommands:
  bar
  zip
$ foo bar --help
Usage of bar:
  -alpha int
  -beta int
  -epsilon int

$ foo bar -alpha 1 -beta 2 -epsilon 3
```


gengokit/template/NAME-service/NAME-cli-client/client_main.gotemplate
```diff
@@ -31,27 +31,47 @@ var (
 
 
 func main() {
+	var (
+		httpAddr       = flag.String("http.addr", "", "HTTP address of addsvc")
+		grpcAddr       = flag.String("grpc.addr", ":5040", "gRPC (HTTP) address of addsvc")
+	)
+
+
```
adamryman: Nit: Whitspace

adamryman: Nit: Whitespace

gengokit/template/NAME-service/NAME-cli-client/client_main.gotemplate
```diff
@@ -31,27 +31,47 @@ var (
...
 
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, "Usage of %s:\n", os.Args[0])
+		flag.PrintDefaults()
+		fmt.Fprintf(os.Stderr, "Subcommands:\n")
+	{{range $meth, $args := .ClientArgs.MethArgs }}
+		fmt.Fprintf(os.Stderr, "  %s\n", "{{ToLower $meth}}")
+	{{end}}
+		fmt.Fprintf(os.Stderr, "\n")
+
```
adamryman: Nit: Whitespace

gengokit/template/NAME-service/NAME-cli-client/client_main.gotemplate
```diff
@@ -31,27 +31,47 @@ var (
...
 	var (
 		{{.ClientArgs.AllFlags}}
 	)
-	flag.Parse()
 
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, "Usage of %s:\n", os.Args[0])
+		flag.PrintDefaults()
+		fmt.Fprintf(os.Stderr, "Subcommands:\n")
+	{{range $meth, $args := .ClientArgs.MethArgs }}
```
adamryman: So that there is not weird spacing in the generated code between these fmt.Fprints's.

```
{{range thing -}}
    //Stuff
{{end -}
```

**More optional:** Also I personally like indenting the generated code blocks as if they were code. Indentation will be fixed by gofmt so it makes it easier to read the template. For me at least, @hasIan has agreed in the past.

gengokit/template/NAME-service/NAME-cli-client/client_main.gotemplate
```diff
@@ -71,33 +91,32 @@ func main() {
...
+			os.Exit(1)
+		}
+		fmt.Println("Client Requested with:")
+		fmt.Println({{$args.CallArgs}})
+		fmt.Println("Server Responded with:")
+		fmt.Println(v)
+{{end}}
 	default:
-		fmt.Fprintf(os.Stderr, "error: invalid method %q\n", method)
+		flag.Parse()
```
adamryman: Why call `flag.Parse()` again? Does this do something?

gengokit/template/NAME-service/NAME-cli-client/client_main.gotemplate
```diff
@@ -31,27 +31,47 @@ var (
...
 
 	var (
 		{{.ClientArgs.AllFlags}}
 	)
-	flag.Parse()
 
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, "Usage of %s:\n", os.Args[0])
+		flag.PrintDefaults()
+		fmt.Fprintf(os.Stderr, "Subcommands:\n")
```
adamryman: Nit, perhaps add a newline here (i.e. `"\nSubcommands:\n"`) as I think the output would be nicer.

Current:
```
$ echo-cli-client
Usage of echo-cli-client:
  -grpc.addr string
        gRPC (HTTP) address of addsvc (default ":5040")
  -http.addr string
        HTTP address of addsvc
Subcommands:
  echo
  louder
$
```
Newline:
```
$ echo-cli-client
Usage of echo-cli-client:
  -grpc.addr string
        gRPC (HTTP) address of addsvc (default ":5040")
  -http.addr string
        HTTP address of addsvc

Subcommands:
  echo
  louder
$ 
```

adamryman (approved):
This is great! We already talked about this yesterday and the improvements are awesome.

Got some nitpicks that should be easy to deal with.

Gonna also do a comment that is out of scope but things we might want to talk about going forward.

adamryman:
Something I have noticed is that we are doing both:

`defer conn.Close` and `os.Exit(1)`, an issue with this is that `os.Exit()` does not cleanup. It terminates. So all those defers will not happen.

One way to deal with this is to pull most of the code out of `main` and have main call a different function which will contain all of the `defer`s and will return exit codes. That way when that function returns all the `defer`s are invoked and we can os.Exit with the code.

adamryman:
Another thing that could be nice is having a global flag for verbosity. To make these cli clients useful for writing a quick script, it would be nice if it could just output the `json`response of the server, or a `json` style error. 

Something like:
```
$ echo-cli-client echo -in Hi
{ "out": "Hi" }

// vs

$ echo-cli-client --verbose echo -in Hi
Client Requested with:
{ "in": "Hi" }
Server Responded with:
{ "out": "Hi" }
// or
...
Server Responded with:
{ "error": "all the bad things and non of the good" }
```

gengokit/template/NAME-service/NAME-cli-client/client_main.gotemplate
hasLeland: That was prior code which should have been changed; thank you and fixing.

gengokit/template/NAME-service/NAME-cli-client/client_main.gotemplate
```diff
@@ -31,27 +31,45 @@ var (
...
-	flag.Parse()
 
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, "Usage of %s:\n", os.Args[0])
+		flag.PrintDefaults()
+		fmt.Fprintf(os.Stderr, "\n")
+		fmt.Fprintf(os.Stderr, "Subcommands:\n")
+		{{range $meth, $args := .ClientArgs.MethArgs -}}
+			fmt.Fprintf(os.Stderr, "  %s\n", "{{ToLower $meth}}")
+		{{end}}
```
adamryman: Missed this `-` after the `end` here. Without it, this line stays in the generated code. In the original comment I said

```
{{range thing -}}
    //Stuff
{{end -}
```

All good, just making that generated code look lean.

PR #155, opened 2017-02-17
Fix #152: Migrate internals to ServiceName instead of PackageName

### LINKED ISSUE #152: Migrate gengokit internals entirely to `ServiceName` instead of `PackageName`
Within `gengokit` there remain a handful of locations where code needlessly makes use of `PackageName` for labeling things, when it should refer to `ServiceName`. For example, within the `server_handler.go` generated file, the name of the service struct is based on `PackageName` when it makes much more sense that it be based on `ServiceName`.

Additionally, the internals of `gengokit/handler` could do with a bit of cleanup in general (whitespace, formatting, clarity, etc), but those cleanliness changes are less important than migrating the names, as the naming inconsistencies make the code needlessly confusing.

hasLeland:
This replaces all un-necessary references to `PackageName` with `ServiceName`, and modifies tests to reflect this change. Additionally, it has some code quality changes, e.g. fixing trailing whitespace.

### DIFF
--- a/gengokit/handler/handler_test.go
+++ b/gengokit/handler/handler_test.go
@@ -50,8 +50,8 @@ func TestServerMethsTempl(t *testing.T) {
 			string output = 1;
 		}
 
-		// ProtoService is a service
-		service ProtoService {
+		// Proto is a service
+		service Proto {
 			// ProtoMethod is simple. Like a gopher.
 			rpc ProtoMethod (RequestMessage) returns (ResponseMessage) {
 				// No {} in path and no body, everything is in the query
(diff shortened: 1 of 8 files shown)


adamryman:
We will want to make sure the service struct in `server_handler.go` is an unexported type. So it needs to be lowercased on the first character. Or an underscore could be added though that is a bit janky.

gengokit/handler/handler_test.go
```diff
@@ -312,17 +314,17 @@ func TestPruneDecls(t *testing.T) {
 
 		// NewService returns a naïve, stateless implementation of Service.
 		func NewService() pb.ProtoServiceServer {
-			return generalService{}
+			return protoserviceService{}
```
adamryman: As a part of cleanup here, it would be good to replace:

`service ProtoService {` with `service Proto {` or something like that. 

As these tests look really weird with `protoserviceService`.

adamryman (approved):
One more thing to clean up in the tests. Though this looks great otherwise!

gengokit/handler/handler_test.go
hasLeland: Done!

adamryman (approved):

PR #156, opened 2017-02-18
Fix #153: exit behavior in generated CLI client preventing defers from firing

### LINKED ISSUE #153: Fix exit behavior in generated CLI client preventing defers from firing
As @adamryman pointed out [on PR #151](https://github.com/TuneLab/go-truss/pull/151#issuecomment-280430848),  within the generated `NAME-cli-client`, we have a main function which sets up `defer`ed functions while also callings `os.Exit(#)` to return error codes, which unfortunately prevents the `defer`ed functions from ever being called.

Adam's suggested solution seems like a good one:

> One way to deal with this is to pull most of the code out of `main` and have `main` call a different function which will contain all of the `defers` and will return exit codes. That way when that function returns, all the defers are invoked and we can call `os.Exit` in the actual `main` with the code returned by the "pseudo-main".

hasLeland:
Moves the "business logic" of the main function for CLI clients into a sub-main function which returns error codes to the actual main function, allowing the deferred functions within the pseudo-main functions to fire while the actual main function then calls `os.Exit`.


gengokit/template/NAME-service/NAME-cli-client/client_main.gotemplate
```diff
@@ -1,121 +1,13 @@
...
-			fmt.Println({{$args.CallArgs}})
-			fmt.Println("Server Responded with:")
-			fmt.Println(v)
-	{{end}}
-	default:
-		flag.Usage()
-		os.Exit(1)
-	}
+	retcode := submain()
+	os.Exit(retcode)
```
adamryman: `os.Exit(submain())`

gengokit/template/NAME-service/NAME-cli-client/client_submain.gotemplate
```diff
@@ -0,0 +1,126 @@
...
+	_ = errors.Wrapf
+	_ = pb.Register{{.Service.Name}}Server
+)
+
+// submain exists to act as the functional main, but will return exit codes to
+// the actual main instead of calling os.Exit directly. This is done to allow
+// the defered functions to be called, since if os.Exit where called directly
+// from this function, none of the defered functions set up by this function
+// would be called.
+func submain() int {
```
adamryman: Why not just called run?

gengokit/template/NAME-service/NAME-cli-client/client_submain.gotemplate
```diff
@@ -0,0 +1,126 @@
+package main
```
adamryman: Why another file?

adamryman (commented):
It looks good, though the names are really weird and verbose.

Take a look at this post https://npf.io/2016/10/reusable-commands/

Also I think this can just be in the same file unless you are actually going to split it into another package.

gengokit/template/NAME-service/NAME-cli-client/client_submain.gotemplate
hasLeland: Calling it `run` seems like using `self` as a method receiver; overly generic, conveying nothing of the purposes this code is meant to fulfill. Nearly any function could accurately be called `run`.

However, it also would violate the convention of a `run` package described in the blog post you linked above which describes a `run` package as (emphasis mine):

> The run package contains the meat of the logic of your binary. You should write this package as if it were a standalone library. *It should be far removed from any thoughts of CLI, flags, etc. It should take in structured data and return errors.*

Since the structure and purpose of the convention we're using in this PR is so narrow and so focused, I think that a name of `main` and `submain` with the explanatory comments is the way to go.

gengokit/template/NAME-service/NAME-cli-client/client_submain.gotemplate
hasLeland: That's a good point; it ended up a separate file during coding, but it definitely makes much more sense for them to live together within the same file.

adamryman (approved):

Issue #154, opened 2017-02-17
Add global verbosity flag for generated CLI clients


hasLeland:
A great suggestion from @adamryman on [PR #151](https://github.com/TuneLab/go-truss/pull/151#issuecomment-280431984) is to add a verbosity flag to the generated CLI client, causing the client to print more information, and certain information, such as the server's response, as JSON output. 

Quoting his suggestion, it might be something like:
```
$ echo-cli-client echo -in Hi
{ "out": "Hi" }

// vs

$ echo-cli-client --verbose echo -in Hi
Client Requested with:
{ "in": "Hi" }
Server Responded with:
{ "out": "Hi" }
// or
...
Server Responded with:
{ "error": "all the bad things and non of the good" }
```

hasLeland:
I implemented this a bit back, but it needs a bit of work. Future me, see the aforementioned PR for the implementation, and this time use functional options!

PR #157, opened 2017-02-18
Implement #154: add client verbosity; make default less verbose

hasLeland:
This PR changes the way the generated CLI will print it's output by default. This change would have the client by default only print the json form of the input and the output, with errors automatically converted to json format. However, the user may tell the client to print more information using the newly added `-verbose` flag.


gengokit/httptransport/templates.go
```diff
@@ -233,6 +237,7 @@ var (
 	_ = ioutil.NopCloser
 	_ = pb.Register{{.Service.Name}}Server
 	_ = io.Copy
+	Verbose = false
```
adamryman: Could this be in its own declaration block with a comment? 

Perhaps a comment could be added to the `_` block and mention it is for compiling even if those packages are not used.

adamryman (commented):
Request:

Could you split up the templates like `gengokit/middlewares/templates` is split up? 

Create a `gengokit/httptransport/templates` directory and have 2 files.

Name them so we know which is for the client and which is the server:

Maybe
`server.go` (which becomes transport_http.go)
`client.go` (which becomes client/http/client.go)

Probably deserves it's own pull request, though I think this one will be easier to follow afterwards. After that gets merged in, you can put this one on top of that one.

gengokit/httptransport/templates.go
hasLeland: That seems like an excellent change, will do!

adamryman:
Looking good! 

I think you are correct that adding this as a global variable feels bad. It is an improvement over sending it through the context though.

Now that the templates are separated some more, and idea came to mind. Why don't we move all the Encoding / Decoding methods that the client uses, to the client package?

That way setting the verbosity could be one of the `ClientOption...` that can be passed in.

The only reason that the methods are in the `transport_http.go` file, is because that is where they existed in the go-kit example.

Also could you move the separating of the templates to a different PR? I want to get that merged in.

PR #159, opened 2017-03-27
Use PBFieldName (json tag name) as fieldname in query string

### LINKED ISSUE #126: Names of attributes within query string
When using the exposed HTTP API of a Truss generated web service the query parameter names are required to be in PascalCase. This makes sense when you think about the underlying Go data structure, but it would feel a bit more HTTPish (this is of course debatable) and consistent if instead the snake cased names used for the JSON attributes were accepted.

zaquestion:
Closes #126 

When using the exposed HTTP API of a Truss generated web service the
query parameter names were required to be in PascalCase. This made sense
thinking about the underlying Go data structure, but not when working
with HTTP / QueryParams. Likewise its more consistent if
instead the snake cased names are used as they will match the json
fields that are accepted.

### DIFF
--- a/gengokit/httptransport/types.go
+++ b/gengokit/httptransport/types.go
@@ -36,6 +36,9 @@ type Binding struct {
 // useful for templating http transport.
 type Field struct {
 	Name string
+	// PBFieldName 'snake_case' in `protobuf:"varint,1,opt,name=snake_case,json=snakeCase" json:"snake_case,omitempty"`
+	// Where Name is 'SnakeCase'
+	PBFieldName string
 	// The name of this field, but passed through the CamelCase function.
 	// Removes underscores, adds camelcase; "client_id" becomes "ClientId".
 	CamelName string
--- a/svcdef/svcdef.go
+++ b/svcdef/svcdef.go
@@ -22,9 +22,8 @@ import (
 	"go/parser"
 	"go/token"
 	"io"
+	"reflect"
 	"strings"
-	"unicode"
-	"unicode/utf8"
 
 	"github.com/pkg/errors"
 )
@@ -75,7 +74,10 @@ type ServiceMethod struct {
 // Field represents a field on a protobuf message.
 type Field struct {
 	Name string
-	Type *FieldType
+	// PBFieldName 'snake_case' in `protobuf:"varint,1,opt,name=snake_case,json=snakeCase" json:"snake_case,omitempty"`
+	// Where Name is 'SnakeCase'
+	PBFieldName string
+	Type        *FieldType
 }
 
 // FieldType contains information about the type of one Field on a message,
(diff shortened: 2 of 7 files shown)


svcdef/svcdef.go
```diff
@@ -415,6 +417,14 @@ func NewField(f *ast.Field) (*Field, error) {
 	// both, and modifying the return value accordingly.
 	var typeFollower func(ast.Expr) error
 	typeFollower = func(e ast.Expr) error {
+		if f.Tag != nil {
+			r := regexp.MustCompile(`json:"(.+),`)
+			m := r.FindStringSubmatch(f.Tag.Value)
+			if len(m) > 1 {
```
adamryman: If `protoc-gen-go` changes what order it writes the struct tags, or if json is used again in the struct tag, this could be brittle using `regexp`.

For example in an old version of `protoc-gen-go` the struct tags looks like this:
```
protobuf:"bytes,32,rep,name=BBB" json:"BBB,omitempty" protobuf_key:"varint,1,opt,name=key" protobuf_val:"fixed32,2,opt,name=value
```
Which is in some of the `svcdef` tests.

Your code parses the `PBFieldName` to be the following:
```
BBB,omitempty" protobuf_key:"varint,1,opt,name=key" protobuf_val:"fixed32,2,opt
```

Which is not desired.

We could use `reflect` to parse the `StructTags`. I looked at `encoding/json` and put this little example together that seems to work.

```
if f.Tag != nil {
	tag := reflect.StructTag(f.Tag.Value).Get("json")
	if idx := strings.Index(tag, ","); idx != -1 {
		rv.PBFieldName = tag[:idx]
	}
}
```

This should avoid the issues of using regex to parse these.

zaquestion:
pesky greedy matches, changed to reflect, good catch!

gengokit/httptransport/types.go
```diff
@@ -36,6 +36,9 @@ type Binding struct {
 // useful for templating http transport.
 type Field struct {
 	Name string
+	// PBFieldName 'snake_case' in `protobuf:"varint,1,opt,name=snake_case,json=snakeCase" json:"snake_case,omitempty"`
+	// Where Name is 'SnakeCase'
+	PBFieldName string
```
adamryman, line 41: If all we are using this for is the query parameter field name, maybe it should just be something like:
```
QueryFieldName
```

svcdef/svcdef.go
```diff
@@ -75,7 +74,10 @@ type ServiceMethod struct {
 // Field represents a field on a protobuf message.
 type Field struct {
 	Name string
-	Type *FieldType
+	// PBFieldName 'snake_case' in `protobuf:"varint,1,opt,name=snake_case,json=snakeCase" json:"snake_case,omitempty"`
+	// Where Name is 'SnakeCase'
+	PBFieldName string
```
adamryman, line 79: How about

```
// PBFieldName is the string value of the field name in the definition .proto file.
// protoc-gen-go stores this value in the json struct tag. See here: 
// https://github.com/golang/protobuf/blob/c9c7427a2a70d2eb3bafa0ab2dc163e45f143317/protoc-gen-go/generator/generator.go#L1785
PBFieldName string

adamryman (approved):
This looks awesome!

A naming suggestion and a possible change to a comment follow below. Nice small diff, cleaning up a few  little things along the way.

zaquestion:
Going to merge as is, since #160 is on top of this. Valid comments I'll see about doing a follow on PR to touch those up.

PR #160, opened 2017-03-28
(decode) Remove unneed dictionary creation, only check error when present

zaquestion:
`QueryParams()` removed resulting generated code changes is thusly
from
```
InEchoStr := queryParams["In"]
InEcho := InEchoStr
// TODO: Better error handling
if err != nil {
	fmt.Printf("Error while extracting InEcho from query: %v\n", err)
	fmt.Printf("queryParams: %v\n", queryParams)
	return nil, err
}
req.In = InEcho

``` 
to
```
queryParams := r.URL.Query()
_ = queryParams

if InEchoStrArr, ok := queryParams["In"]; ok {
	InEchoStr := InEchoStrArr[0]
	InEcho := InEchoStr
	req.In = InEcho
}
```

As an added bonus, queryParams will no longer override body parameters when
not present

Additionally the error logic in the top block was always generated even if
the InEchoStr was of type string and thus passed through directly
(did not need to be parsed with ex. ParseBool). The code was unreachable
and will now nolonger be generated


gengokit/httptransport/embeddable_funcs.go
```diff
@@ -70,19 +69,3 @@ func RemoveBraces(val string) string {
 	val = strings.Replace(val, "}", "", -1)
 	return val
 }
-
-// QueryParams takes query parameters in the form of url.Values, and returns a
```
adamryman, line 74: Yeah, kill that code.

adamryman (commented):
Amazing great job!

Don't really like that it is on top of #159, though I can see why as you touch the same file and a couple of the same lines. 

Seeing it here, I really think in `httptransport.go`, `PBFieldName` should be `QueryFieldName` or something similar, as it makes a lot more sense reading it in the Go template code. So I think that change should be made in #159 and reflected here.

Seems like a simpler way to do it, and the tests pass on 1.7 ci, and 1.8 my machine.

gengokit/httptransport/httptransport.go
```diff
@@ -246,6 +259,7 @@ func createDecodeConvertFunc(f Field) string {
 		fType = "%s, err := strconv.ParseFloat(%s, 64)"
 	case strings.Contains(f.GoType, "string"):
```
hasIan, line 260: Why is this `strings.Contains` instead of `==`? Is there a specific reason for this wider match that I am missing?

gengokit/httptransport/templates_test.go
```diff
@@ -173,12 +175,9 @@ func DecodeHTTPSumZeroRequest(_ context.Context, r *http.Request) (interface{},
 		fmt.Printf("Error while reading path params: %v\n", err)
 		return nil, errors.Wrap(err, "couldn't unmarshal path parameters")
 	}
-	queryParams, err := QueryParams(r.URL.Query())
+
+	queryParams := r.URL.Query()
```
hasIan, line 179: Is assigning this to a variable really even necessary? It was with the old `QueryParams` function, but seems not so now.

hasIan (approved):
Looks good. Clean. 👍  :100: :shipit:

gengokit/httptransport/templates_test.go
zaquestion: Pretty necessary, that line is part of a larger template for the DecodeHTTPEndpointRequest.

The template for the individual queryparams relies on `queryParams` existing. Otherwise I'd need to use r.URL.Query() for every queryparam

hasIan: > Otherwise I'd need to use r.URL.Query() for every queryparam

Right. So?

gengokit/httptransport/httptransport.go
hasIan: Any answer here? Curiosity compels me.

Issue #158, opened 2017-03-26
Middlewares: WrapAllExcept


adamryman:
The ability to exclude `Status` and `Ping` endpoints from being wrapped in middlewares has been requested. Specifically, to wrap all endpoints except `Status` and `Ping` with authentication middleware for `opsgenie`. 

Proposal: 
```
func (*e Endpoints) WrapAllExcept(middlware endpoint.Middleware, except ...endpoint.Endpoint) {
```

This function would allow you to wrap all endpoints in a middleware and specify the ones you do not want.

```
in.WrapAllExcept(authMiddleware, in.Status, in.Ping)
```

adamryman:
Or `WrapAll` could be modified from

```
func (e *Endpoints) WrapAll(middlewares ...endpoint.Middleware) {
```
To
```
func (e *Endpoints) WrapAll(middleware endpoint.Middleware, except ...endpoint.Endpoint) {
```

And to wrap all endpoints in multiple middlewares, `WrapAll` just needs to be called multiples times.

```
in.WrapAll(authMiddleware, in.Status, in.Ping)
in.WrapAll(loggingMiddleware)
in.WrapAll(metricsMiddleware)
```

zaquestion:
Infavor of making modifying `WrapAll`, We could make it an array of `endpoint.Middleware`

hasIan:
I also think there's merit for auto-generating Status and Ping handlers as part of Truss generation. I mean, we're generating services that are intended to sit behind a load balancer; when aren't you going to want a status and/or ping handler is, the answer is, "Never. You will never not want these things." We have them listed in every `service.proto` we have made.

I can see an argument for `Status` to be something different, but I can imagine a generalized response that indicates a general status and includes an optional list of messages to clarify why a status may not be `STELLAR` or otherwise indicating that the service is free of trouble.

hasIan:
Note that I still think a `WrapAllExcept` or whatnot is a good idea, even considering my most recent comment.

adamryman:
Implemented in #164

PR #164, opened 2017-03-28
WrapAll(middlewares, exclude ...string)

adamryman:
Add exclude functionality to `endpoints.WrapAll` as described in #158.

With integration tests, which were important, because my first implementation in fact did have bugs.

```
// WrapAll wraps each Endpoint field of struct Endpoints with a
// go-kit/kit/endpoint.Middleware.
// Use this for applying a set of middlewares to every endpoint in the service.
// Optionally, endpoints can be passed in by name to be excluded from being wrapped.
// WrapAll(middleware, "Status", "Ping")
func (e *Endpoints) WrapAll(middleware endpoint.Middleware, excluded ...string) {
```
```
// Pass in the middlewares you want applied to every endpoint.
// optionally pass in endpoints by name that you want to be excluded
// e.g.
// in.WrapAll(authMiddleware, "Status", "Ping")
```


adamryman:
Note that must of the lines added are to the test harness for the middlewares integration tests.

See this file for the actual implementation. 

https://github.com/TuneLab/go-truss/pull/164/files#diff-72366305828c07ae0e996c1ae3b5674c

gengokit/template/NAME-service/generated/endpoints.gotemplate
```diff
@@ -63,14 +63,24 @@ type Endpoints struct {
...
+// WrapAll(middleware, "Status", "Ping")
+func (e *Endpoints) WrapAll(middleware endpoint.Middleware, excluded ...string) {
+	included := map[string]struct{}{
+		{{- range $i := .Service.Methods}}
+			"{{$i.Name}}": struct{}{},
+		{{- end}}
+	}
+
+	for _, ex := range excluded {
+		delete(included, ex)
```
zaquestion: I'm thinking we should probably log an error (fatal?) if the middleware is not one of the candidates i.e. spelling error.

hasAdamr: We would have to check for that, I could generate a code block for it though.

hasAdamr: We also don't have a logger here, not sure how to bring one into to this file either.

hasIan: I have a few thoughts here: 

1. Yes, this should be a fatal error if we cannot internally reconcile an endpoint with its name. I don't even know what sort of situation would lead to this, since the user is passing in endpoints, and not names of endpoints.
2. Since using Truss implies that you are using Go-Kit, it is okay to use Go-Kit's logger in the generated code.
3. If the error in this case is indicative of a programming error such that the service isn't going to get out the door without fixing it, I think this is an okay place to use the builtin `panic`. You can avoid thinking about the log altogether.

hasAdamr: @ian the user is passing in Endpoints by name. There was no good way to get the struct field name, it is lost in the function call. I could explain in person, though there are a lot of particulars. Zaq and I ran through the options this morning.

We are using panic now.

zaquestion (approved):

PR #169, opened 2017-03-31
Remove redundant package aliases from generated imports

hasLeland:
Since the packages named 'svc' and 'middlewares' are actually in directories with those names now, we no longer need to alias them to those names.

- [x] Remove redundant 'svc' package alias within:
    - [X] `cmd/NAME-server/main.gotemplate`
    - [X] `svc/client/http/client.gotemplate`
    - [X] `svc/client/grpc/client.gotemplate`
    - [X] `middlewares/endpoints.gotemplate`
- [X] Remove redundant 'middlewares' aliases within:
    - [X] `cmd/NAME-server/main.gotemplate`
- [X] Remove redundant 'handler' aliases in the:
    - [X] `cmd/NAME-server/main.gotemplate`, also change references to 'handlers'


gengokit/template/NAME-service/cmd/NAME-server/main.gotemplate
```diff
@@ -21,9 +21,9 @@ import (
 
 	// This Service
 	handler "{{.ImportPath -}} /handlers"
```
adamryman: `handler` should be removed as well, and all places that currently call `handler` should be updated to `handlers`. Could you roll that into this pull?

adamryman: Also anywhere else `handler` is used as well.

hasLeland: Done!

adamryman:
How about removing `clientHandler` in `gengokit/template/NAME-service/cmd/NAME/main.gotemplate` and updating all calls to it `handlers`? Thanks

The package should also be updated to `package handlers` rather than `package clienthandler`

adamryman (approved):

PR #170, opened 2017-03-31
Seperate httptransport templates

hasLeland:
- [X] Move `httptransport` template files to a sub-package split into logical files


adamryman:
`gengokit/httptransport/templates.go` should be deleted I believe.

Also should `gengokit/httptransport/templates_test.go` be moved to the `gengokit/httptransport/templates` dir? More unsure about this one.

adamryman (approved):

hasLeland:
@adamryman With the current tests in `templates_test.go`, that's not really possible, since they test the *generation* of tests, which requires that it use functionality within the `httptransport` package. If it where placed within `templates/`, it creates an import cycle.

`templates_test.go` could be renamed or split up, but that's a different thing (I think, at least).

PR #171, opened 2017-03-31
Remove automatic server-side prints

hasLeland:
Remove print statements which are automatically run by the server, so errors and such will *not* print to the stdout of the server, instead they'll be included in the response.

### DIFF
--- a/gengokit/httptransport/templates/server.go
+++ b/gengokit/httptransport/templates/server.go
@@ -18,7 +18,6 @@ var ServerDecodeTemplate = `
 		pathParams, err := PathParams(r.URL.Path, "{{$binding.PathTemplate}}")
 		_ = pathParams
 		if err != nil {
-			fmt.Printf("Error while reading path params: %v\n", err)
 			return nil, errors.Wrap(err, "couldn't unmarshal path parameters")
 		}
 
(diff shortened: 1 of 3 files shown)


gengokit/httptransport/templates/server.go
```diff
@@ -262,4 +250,16 @@ func headersToContext(ctx context.Context, r *http.Request) context.Context {
...
+	return func(ctx context.Context, r *http.Request) (interface{}, error) {
+		logger.Log("method", r.Method, "url", r.URL.String())
+		rv, err := next(ctx, r)
+		if err != nil {
+			logger.Log("method", r.Method, "url", r.URL.String(), "Error", err)
+		}
+		return rv, err
+	}
+}
+
```
adamryman: Trailing whitespace

adamryman (approved):
One nit. Perfect otherwise!

gengokit/httptransport/templates/server.go
hasLeland: Done!

zaquestion (approved):

gengokit/httptransport/templates/server.go
```diff
@@ -262,4 +250,15 @@ func headersToContext(ctx context.Context, r *http.Request) context.Context {
 
 	return ctx
 }
+
+func HTTPDecodeLogger(next httptransport.DecodeRequestFunc, logger log.Logger) httptransport.DecodeRequestFunc {
```
zaquestion, line 254: Any reason for keeping?

hasLeland: Basically, as a "just in case", but that may or may not be valid. Think it should be deleted?

PR #172, opened 2017-04-01
Split -server/main.go into a few component packages

adamryman:
Moved things out of main, getting us closer to be purely configure by import.

Not meant to be a final solution, just an incremental step in the right direction.

Also: 
- Move cli package to client, as it is the client/cli
- Move TestPortVariable to be first test to run as it is the fastest test


gengokit/template/NAME-service/cmd/NAME-server/main.gotemplate
```diff
@@ -57,82 +27,5 @@ func main() {
...
-		s := grpc.NewServer()
-		pb.Register{{.Service.Name}}Server(s, srv)
-
-		logger.Log("addr", *grpcAddr)
-		errc <- s.Serve(ln)
-	}()
-
-	// Run!
-	logger.Log("exit", <-errc)
+	server.Run(cli.Config, logger)
```
hasLeland, line 30: :+1:

hasLeland (approved):

PR #174, opened 2017-04-08
Fix for RPC camelcase binding

hasLeland:
Fixes a bug where RPC methods with names that changed when camelcased would fail to have HTTP transport generated, and all HTTP transport encode/decode functions thereafter would also fail to be generated.

This was because camelcased names where improperly handled within `svcdef.assembleHTTPParams`, and errors returned from that function wheren't being checked : (

But now that's fixed, and a test has been added!


adamryman (approved):
Looks good! :+1:

PR #176, opened 2017-04-10
Parse and convert enums from query string parameters

zaquestion:
- Added integration tests

### DIFF
--- a/gengokit/httptransport/httptransport.go
+++ b/gengokit/httptransport/httptransport.go
@@ -107,10 +107,17 @@ func NewBinding(i int, meth *svcdef.ServiceMethod) *Binding {
 			newField.GoType = "[]" + newField.GoType
 		}
 
+		// IsEnum needed for ConvertFunc and TypeConversion logic just below
+		newField.IsEnum = field.Type.Enum != nil
 		newField.ConvertFunc, newField.ConvertFuncNeedsErrorCheck = createDecodeConvertFunc(newField)
 		newField.TypeConversion = createDecodeTypeConversion(newField)
 
 		nBinding.Fields = append(nBinding.Fields, &newField)
+
+		// Enums are allowed in query/path parameters, skip warning
+		if newField.IsEnum {
+			continue
+		}
 
 		// Emit warnings for certain cases
 		if !newField.IsBaseType && newField.Location != "body" {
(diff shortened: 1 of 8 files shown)


gengokit/httptransport/httptransport.go
```diff
@@ -107,11 +107,18 @@ func NewBinding(i int, meth *svcdef.ServiceMethod) *Binding {
...
 		}
 
+		// IsEnum needed for ConvertFunc and TypeConversion logic just below
+		newField.IsEnum = field.Type.Enum != nil
 		newField.ConvertFunc, newField.ConvertFuncNeedsErrorCheck = createDecodeConvertFunc(newField)
 		newField.TypeConversion = createDecodeTypeConversion(newField)
 
 		nBinding.Fields = append(nBinding.Fields, &newField)
 
+		// Enums are allowed in query/path parameters, skip warning
```
adamryman, line 117: If Enums are allowed as both query and path parameters, we should have a test for the path parameters.

I see a test for use as query parameter, though not one for the path parameter.

We don't need to support it in the path currently, though unless we have a test we should warn about it.

gengokit/httptransport/httptransport.go
```diff
@@ -241,25 +247,31 @@ req.{{.CamelName}} = {{.TypeConversion}}
 func createDecodeConvertFunc(f Field) (string, bool) {
 	needsErrorCheck := true
 	fType := ""
-	switch {
-	case strings.Contains(f.GoType, "uint32"):
+	switch f.GoType {
```
adamryman, line 262: Very nice cleanup :+1:

adamryman (commented):
Looks good! Though we need a test for Enums in the path if we are no longer warning about it. 

Looks like the query is working though!

gengokit/httptransport/httptransport.go
```diff
@@ -319,16 +343,15 @@ func createDecodeTypeConversion(f Field) string {
...
-	switch {
-	case strings.Contains(f.GoType, "uint32"):
-		fType = "uint32(%s)"
-	case strings.Contains(f.GoType, "int32"):
-		fType = "int32(%s)"
-	case strings.Contains(f.GoType, "float32"):
-		fType = "float32(%s)"
+	switch f.GoType {
+	case "uint32", "int32", "float32":
+		fType = f.GoType + "(%s)"
```
hasLeland, line 348: :arrow_up: :exclamation: :tada:

hasLeland (approved):

PR #179, opened 2017-04-18
panic when a user tries to create an HTTP client if there are no http endpoints. (client will panic later in a non-obvious way otherwise)

zaquestion:


adamryman (approved):
LGTM

PR #180, opened 2017-04-26
Set request-url and transport type in http transport to allow the information to be used in middlewares

zaquestion:


gengokit/httptransport/templates_test.go
```diff
@@ -69,6 +69,9 @@ func EncodeHTTPSumZeroRequest(_ context.Context, r *http.Request, request interf
 	req := request.(*pb.SumRequest)
 	_ = req
 
+	r.Header.Set("transport", "HTTPJSON")
```
hasIan, line 72: It's not clear why this is being done in the test, when the exact same thing is being done in the generated client. I'm sure there's a reasonable explanation, but it does not present itself.

cmd/_integration-tests/transport/handlers/handlers.go
```diff
@@ -20,6 +21,14 @@ type transportpermutationsService struct{}
 
 // GetWithQuery implements Service.
 func (s transportpermutationsService) GetWithQuery(ctx context.Context, in *pb.GetWithQueryRequest) (*pb.GetWithQueryResponse, error) {
+	spew.Dump(ctx)
+	if rurl := ctx.Value("request-url").(string); rurl != "/getwithquery" {
+		panic("Context Value: request-url, expected 'GetWithQuery' got " + rurl)
```
hasIan: The string in this panic does not match the value that's actually being tested against. It's an integration test, so it's not a critical path anything, but it's an inconsistency nonetheless.

cmd/_integration-tests/transport/handlers/handlers.go
```diff
@@ -20,6 +21,14 @@ type transportpermutationsService struct{}
 
 // GetWithQuery implements Service.
 func (s transportpermutationsService) GetWithQuery(ctx context.Context, in *pb.GetWithQueryRequest) (*pb.GetWithQueryResponse, error) {
+	spew.Dump(ctx)
```
hasAdamr: Not sure if spew should be in this handler.

cmd/_integration-tests/transport/handlers/handlers.go
```diff
@@ -20,6 +21,14 @@ type transportpermutationsService struct{}
...
+	if rurl := ctx.Value("request-url").(string); rurl != "/getwithquery" {
```
hasAdamr: This is not really a test. It panics inside the handlers rather than asserting on values in the `http_test.go` file.

See test [TestCtxToCtxViaHTTPHeaderClient](https://github.com/adamryman/go-truss/blob/d1d3f0016e9b975fd8e080182792744e6b6d206a/cmd/_integration-tests/transport/http_test.go#L168) for a header test example.

hasIan (approved):
It looks as though we're polluting Truss with leaked implementation details from our pre-Truss services here, but since we are the main consumers of Truss, and this isn't going to hurt anything, seems okay to me.

gengokit/httptransport/templates/client.go
```diff
@@ -13,6 +13,9 @@ var ClientEncodeTemplate = `
 		req := request.(*pb.{{GoName $binding.Parent.RequestType}})
 		_ = req
 
+		r.Header.Set("transport", "HTTPJSON")
```
hasAdamr, line 16: Possibly add a comment about why this is added. Though the blame may be enough.

hasAdamr (commented):
Polluting truss with our requirements was bound to happen at some point. Though we can always remove it when we find a better solution.

Looks ok, would like the test to be more standard though.

zaquestion:
There are no requirements being imposed on a consumer of truss. This change adds additional information that is useful on its own. Granted, this change is being made because TUNE needs it.

PR #184, opened 2017-05-22
Add ability to send headers for truss cli clients

hasLeland:
This PR adds logic to the generated truss clients allowing a user to send arbitrary text in the header via the `--header` flag. Example:

```
cli-client --header "nameOfHeader: valueOfHeader" methodname -field var
```

Though the flag is the singular 'header', you can attach multiple headers to a request with repeated `header` flags. For example:

    cli-client --header "first: foo" --header "second: bar" method -field val


gengokit/template/NAME-service/cmd/NAME/main.gotemplate
```diff
@@ -37,12 +37,30 @@ func main() {
 	os.Exit(submain())
 }
 
+type headerSeries []string
+
+func (h *headerSeries) Set(val string) error {
+	requiredPortions := 2
+	parts := strings.SplitN(val, ":", requiredPortions)
+	if len(parts) != requiredPortions {
+		return errors.New(fmt.Sprintf("value %q cannot be split in two; must contain at least one ':' character", val))
```
hasIan: Just use `fmt.Errorf`. It does exactly what you have here, more succinctly.

gengokit/template/NAME-service/cmd/NAME/main.gotemplate
```diff
@@ -37,12 +37,30 @@ func main() {
 	os.Exit(submain())
 }
 
+type headerSeries []string
+
+func (h *headerSeries) Set(val string) error {
+	requiredPortions := 2
```
hasIan: This makes me think I'm going to be eating something. :) I'd just keep calling it parts like you do in the other bits that deal with this. Additionally, it'd be okay to declare this as a `const`.

gengokit/template/NAME-service/cmd/NAME/main.gotemplate
```diff
@@ -37,12 +37,30 @@ func main() {
...
+	parts := strings.SplitN(val, ":", requiredPortions)
+	if len(parts) != requiredPortions {
+		return errors.New(fmt.Sprintf("value %q cannot be split in two; must contain at least one ':' character", val))
+	}
+	*h = append(*h, parts...)
```
hasIan: Probably want to `strings.TrimSpace(parts[1])` since the arguments are likely to be passed `-H "Header: Value"`.

gengokit/template/NAME-service/cmd/NAME/main.gotemplate
```diff
@@ -37,12 +37,30 @@ func main() {
...
+}
+
 // submain exists to act as the functional main, but will return exit codes to
 // the actual main instead of calling os.Exit directly. This is done to allow
 // the defered functions to be called, since if os.Exit where called directly
 // from this function, none of the defered functions set up by this function
 // would be called.
 func submain() int {
+	var headers headerSeries
+	flag.Var(&headers, "header", "Header(s) to be sent in the transport (follows cURL style)")
```
hasIan, line 64: Maybe make this a single-char option like curl has with `-H`, since `-header` would be bothersome to type multiple times.

The best would be [`pflag`](https://github.com/spf13/pflag), which has support for long and short flags, but using this in clients would mean that the user has to have pflag installed locally to build the clients, which is a little too unfriendly of a jump to make, I think.

gengokit/template/NAME-service/cmd/NAME/main.gotemplate
```diff
@@ -37,12 +37,30 @@ func main() {
 	os.Exit(submain())
 }
 
+type headerSeries []string
```
hasIan, line 40: This seems less intuitive to me than a `map[string][]string`, but that's picking nits, I think, because this is a straightforward enough implementation.

gengokit/template/NAME-service/cmd/NAME/main.gotemplate
```diff
@@ -81,17 +99,21 @@ func submain() int {
 		service pb.{{GoName .Service.Name}}Server
 		err     error
 	)
+	ctx := context.Background()
```
hasIan: This doesn't seem to be used in the next couple blocks that initialize the HTTP/GRPC clients. Maybe move this initialization to right before the `switch` where it's used in the `CtxValuesToSend` call, for greater delcaration-usage locality. 

And off-topic, but I can't type that function name without thinking it should just be `ContextValuesToSend`. :)

hasIan (approved):
Looks good--a couple comments for your consideration.

gengokit/template/NAME-service/cmd/NAME/main.gotemplate
hasLeland: You're right that `pflag` is a bit out of scope for this, and given the need to pick either short or long command line parameters, I'll take longer ones every time, as their easier to search for in history and easier to read.

gengokit/template/NAME-service/cmd/NAME/main.gotemplate
```diff
@@ -101,6 +119,15 @@ func submain() int {
...
+		flag.Usage()
+		return 1
+	}
+
+	ctx := context.Background()
+	for i := 0; i < len(headers); i += 2 {
+		ctx = context.WithValue(ctx, headers[i], headers[i+1])
+	}
 
 	switch flag.Args()[0] {
```
hasAdamr, line 134: It would be great to switch this out with
```
args := flag.Args()
if len(args) < 1 {
    // print message about needing method
    os.Exit(1)
}

So it does not panic if flags are provided but not arguments.

hasAdamr (approved):
Seems to function correctly. Great job!

gengokit/template/NAME-service/cmd/NAME/main.gotemplate
hasLeland: I believe that I already added that code, on line 122. Does that code cover what you mentioned in this comment?

gengokit/template/NAME-service/cmd/NAME/main.gotemplate
hasIan: Whereas I will take the shorter one every time. I grew up with DOS. Long command line args? Bloated gilding! :P

PR #186, opened 2017-05-26
Add truss version and commit date to generated files. 

adamryman:
- [X] Update build date to commit date, so that people who build on the same commit to have the same generated files
    - [X] gnu `date` and bsd (mac) `date` compatible script for getting commit date in utc. `commit_date.sh`
- [X] Update generated files with `Version: ` and `Version Date: `
- [X] Update main to prompt user with instructions if truss was not built with make
    - [X] Prompt user if they want truss to reinstall itself with make, then pass through arguments to new truss installation.
- [X] Update `README.md` to install with make


cmd/truss/main.go
```diff
@@ -411,3 +416,57 @@ func fileExists(path string) bool {
...
+
+// promptNoMake prints that truss was not built with make and prompts the user
+// asking if they would like for this process to be automated
+// returns true if yes, false if not.
+func promptNoMake() bool {
+	const msg = `
+truss was not built using Makefile.
+Please run 'make' inside go import path %s.
+
+Do you want to automatically run 'make' and rerun command:
```
zaquestion, line 428: :100:

zaquestion (approved):

PR #187, opened 2017-05-29
Move example LabeledMiddleware out of generated code

adamryman:
Also normalizing on using the word "endpoint" to match gokit documentation.

Also added RequestLatency LabeledMiddleware example.


_example/middlewares/labeledmiddlewares.go
```diff
@@ -0,0 +1,42 @@
...
+				errCount.With("endpoint", name).Add(1)
+			}
+			return resp, err
+		}
+	}
+}
+
+// RequestLatency is a LabeledMiddleware, reporting the request time of and
+// endpoint along with its name
+func RequestLatency(h metrics.Histogram) LabeledMiddleware {
```
zaquestion: Is it implied that this is a for Requests since its a middleware. It seems like every middleware could be prefixed with "Request"?

zaquestion (approved):

PR #189, opened 2017-06-01
Disable HTML escaping in JSON encoder

hasLeland:
The Go standard JSON encoder will escape HTML characters that it encounters while encoding a JSON object. That's dumb, and it messes with responses from Truss, so this PR disables HTML escaping in generated JSON response encoders.

- [x] Disable HTML escaping in generated JSON response encoders
- [x] Modify tests to match change in generated code

### DIFF
--- a/gengokit/httptransport/templates/client.go
+++ b/gengokit/httptransport/templates/client.go
@@ -60,7 +60,9 @@ var ClientEncodeTemplate = `
 				toRet.{{$field.CamelName}} = req.{{$field.CamelName}}
 			{{end}}
 		{{- end }}
-		if err := json.NewEncoder(&buf).Encode(toRet); err != nil {
+		encoder := json.NewEncoder(&buf)
+		encoder.SetEscapeHTML(false)
+		if err := encoder.Encode(toRet); err != nil {
 			return errors.Wrapf(err, "couldn't encode body as json %v", toRet)
 		}
 		r.Body = ioutil.NopCloser(&buf)
--- a/gengokit/httptransport/templates/server.go
+++ b/gengokit/httptransport/templates/server.go
@@ -201,7 +201,9 @@ type errorWrapper struct {
 // EncodeHTTPGenericResponse is a transport/http.EncodeResponseFunc that encodes
 // the response as JSON to the response writer. Primarily useful in a server.
 func EncodeHTTPGenericResponse(_ context.Context, w http.ResponseWriter, response interface{}) error {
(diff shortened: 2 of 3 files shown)


hasAdamr:
I only see a change for the client. Does the server still send out HTML escaped json?

Could we have a test?

hasIan:
Yeah, there's this bit in the service:
```
gengokit/httptransport/templates/server.go
187:	json.NewEncoder(w).Encode(errorWrapper{Error: msg})
204:	return json.NewEncoder(w).Encode(response)
```
In fact, the issue was specifically about this--but, it makes sense to me to disable it in the clients as well. 👍

hasLeland:
@hasAdamr @hasIan I've gone back and put the correct HTML escape-disabling code in. I additionally tried to create a test for this, but from my brief stab at it I couldn't find a nice way to fit a test for this into our existing integration test framework. The issue is that since `json.Unmarshal` will un-escape the escaped HTML, none of our endpoints can see that anything's changed.

The most minimal way to write a test would be to add some kind of transport-level middleware to the transport integration tests that maybe puts the raw bytes into the context for the one endpoint used to test the service, then have the endpoint do some kind of further escaping of that data in it's response so the test body can check if what the middleware saw in the bytes of the body was what we expected. 

But that seems a bit kludg-ie and we don't have a good way to make that happen as of right now. For that reason, I'm going to omit it from this PR, though it's something to think about.

----------------------------
However, I have done some manual testing with `echo-service` and `curl` to ensure that yes, the lines I added most recently do correctly disable the server responding with HTML escaped.

zaquestion:
:100:

hasAdamr:
Should we do this one? https://github.com/hasLeland/go-truss/blob/b2942219bed5db2007830f10513fc4dbd64b30b2/gengokit/httptransport/templates/server.go#L187

hasAdamr:
Should only be for error messages, so I think that is fine if that one is HTML escaped.

Otherwise LGTM

adamryman (approved):

hasLeland:
@hasAdamr Yeah, since that one was only for error messages I figured we can deal with HTML being escaped.

hasIan:
I understand the desire to be paranoid in one's own dependencies, but here, a test for this is basically a test of SetEscapeHTML, and requires too much unnatural effort to shoehorn in. It is safe to say that if we call SetEscapeHTML(false) that the Go stdlib json package respects this. I don't think we've introduced any risk with this change, especially if you've manually vetted the changes.

PR #190, opened 2017-06-13
[MVS-117] Make errors more useful when receiving non-json requests/responses.

adamryman:
Both truss servers and clients would just return unhelpful json decoding errors when decoding requests and responses respectively. 

This PR has both put the body of the request/response into the error message if their body is not json. This is limited to 8KB so that if binary data is sent, these error messages will not be way too massive. 

- [X] Update client to put response body in error if it is not a json response
    - [x] Test that response body is in the error
    - [X] Limit to 8KB (8196 bytes)
        - [x] Test with response body bigger than 8196 bytes.
    - [X] Update error to have status code if it was not `200`
- [X] Update server to put request body in error if the body is not json
    - [x] Test that request body is in the error
    - [X] Limit to 8KB (8196 bytes)
        - [x] Test with request body bigger than 8196 bytes.
- [X] Change server http response content type to `"application/json; charset=utf-8"`
    - [x] Test http response content type

### DIFF
--- a/cmd/_integration-tests/transport/http_test.go
+++ b/cmd/_integration-tests/transport/http_test.go
@@ -24,6 +24,7 @@ var httpAddr string
 var nonJSONHTTPAddr string
 
 const brokenHTTPResponse = `<html> Not json </html>`
+const brokenHTTPRequest = brokenHTTPResponse
 
 func TestGetWithQueryClient(t *testing.T) {
 	var req pb.GetWithQueryRequest
@@ -403,13 +404,9 @@ func TestStrangeRPCName(t *testing.T) {
 	}
 }
 
-// TODO: TestNonJSONRequestBodyIsInError()
-
-// TODO: TestNonJSONRequestBodyIsLessThan8KB()
-
-// Test that if a truss client recieves a non-json response from a "truss"
-// server, that we put that response in the error message, for easier debugging
-// Rather than just getting a json paring error
+// Test that if a truss client receives a non-json response from a "truss"
+// server, that we put that response body in the error message. To allow for
+// developers to see the request body in the errors.
 func TestNonJSONResponseBodyFromClientCallIsInError(t *testing.T) {
 	svchttp, err := httpclient.New(nonJSONHTTPAddr)
 	if err != nil {
@@ -428,7 +425,7 @@ func TestNonJSONResponseBodyFromClientCallIsInError(t *testing.T) {
 }
 
 // Test that if a non json response is recieved that is greater than 8KB, that
-// we only put the first 8KB error, as to not flood our logs
+// we only put the first 8KB error, as to not flood memory with huge errors.
 func TestNonJSONResponseBodyFromClientCallIsLessThan8KB(t *testing.T) {
 	svchttp, err := httpclient.New(nonJSONHTTPAddr)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


cmd/_integration-tests/transport/http_test.go
```diff
@@ -337,18 +341,18 @@ func TestGetWithPathParamsRequest_IncompletePath(t *testing.T) {
...
 	var resp map[string]string
 	err = json.Unmarshal(respBytes, &resp)
 	if err != nil {
-		t.Fatalf("couldn't unmarshal bytes: %v", err)
+		t.Fatalf("cannot unmarshal bytes: %s", respBytes)
 	}
 
 	want := map[string]string{
-		"error": "couldn't unmarshal path parameters: Expected a path containing 4 parts, provided path contains 3 parts",
+		"error": "cannot unmarshal path parameters: expected a path containing 4 parts, provided path contains 3 parts",
```
hasLeland: If we're updating things to follow the present tense, then "expected" should be changed to "expects" or "expecting". Though that would mean many changes throughout and that may or may not be a priority.

cmd/_integration-tests/transport/http_test.go
```diff
@@ -400,6 +404,124 @@ func TestStrangeRPCName(t *testing.T) {
...
+		t.Fatalf("Expected error to be less than 8KB with a little padding, actual %d", l)
+	}
+	t.Log("Non JSON response length", l)
+}
+
+// Test that if a truss server receives a non-json request, we put that request
+// body in the error message. To allow for developers to see the request body in the errors.
+func TestNonJSONRequestBodyIsInError(t *testing.T) {
+	// Put some bad data into the body
+	req, err := http.NewRequest("GET", httpAddr+"/error/non/json", strings.NewReader(brokenHTTPRequest))
```
hasLeland: The semantics for GET requests are pretty explicitly spelled out in [the RFC for HTTP 1.1 as being intended to retrieve data specified in the URI of a request](https://tools.ietf.org/html/rfc2616#section-9.3). So while you *can* put a body into a GET request, it's unusual to do so. It makes much more sense for these to be POST requests and for the associated RPCs in the `.proto` file to serve data from a `post` as well.

cmd/_integration-tests/transport/http_test.go
```diff
@@ -400,6 +404,124 @@ func TestStrangeRPCName(t *testing.T) {
...
+func TestResponseContentType(t *testing.T) {
+	req, err := http.NewRequest("GET", httpAddr+"/content/type", strings.NewReader(""))
+	if err != nil {
+		t.Fatal(errors.Wrap(err, "cannot construct http request"))
+	}
+
+	client := &http.Client{}
+	httpResp, err := client.Do(req)
+	if err != nil {
+		t.Fatal(errors.Wrap(err, "cannot make end http request"))
```
hasLeland: Is this message supposed to say "cannot make end http request"? If so, could you elaborate on what it means? : )

gengokit/httptransport/templates/server.go
```diff
@@ -9,26 +9,34 @@ var ServerDecodeTemplate = `
 	// body. Primarily useful in a server.
 	func DecodeHTTP{{$binding.Label}}Request(_ context.Context, r *http.Request) (interface{}, error) {
 		var req pb.{{GoName $binding.Parent.RequestType}}
-		err := json.NewDecoder(r.Body).Decode(&req)
-		// err = io.EOF if r.Body was empty
-		if err != nil && err != io.EOF {
-			return nil, errors.Wrap(err, "decoding body of http request")
+		buf, err := ioutil.ReadAll(r.Body)
```
hasLeland, line 12: This is good, cleaning up that old gross `json.NewDecoder().Decode()` stuff is a good move. :+1:

hasLeland (changes requested):
Pretty much seems great! Nice to clean things up, and it makes our stuff more useful!

hasLeland (approved):

gengokit/httptransport/templates/client.go
```diff
@@ -212,12 +212,25 @@ func contextValuesToHttpHeaders(keys []string) httptransport.RequestFunc {
 	// error and attempt to decode the specific error message from the response
 	// body. Primarily useful in a client.
 	func DecodeHTTP{{$method.Name}}Response(_ context.Context, r *http.Response) (interface{}, error) {
+		buf, err := ioutil.ReadAll(r.Body)
+		if len(buf) == 0 {
+			return nil, errors.New("response http body empty")
+		}
+
+		if err != nil {
```
hasIan: I might switch the order of these checks, to check the error first. It'll be caught either way--but I don't know for certain what's returned for the buffer if an error occurs during a read: if a zero-byte buffer comes back but there was an actual error, it'll get reported as a zero-length response, and the actual error will not be reported.

gengokit/httptransport/templates/server.go
```diff
@@ -188,7 +197,7 @@ func errorEncoder(_ context.Context, err error, w http.ResponseWriter) {
 }
 
 type errorWrapper struct {
-	Error string ` + "`json:\"error\"`\n" + `
+	Error string ` + "`" + `json:"error"` + "`" + `
```
hasIan, line 200: What is this struct-tag-looking construct? This is a template file, but I still don't understand this, or why it could not just be
```
Error string `json:"error"`
```

hasIan (approved):
Looks good; a couple minor comments to review before merging, perhaps.

gengokit/httptransport/templates/client.go
adamryman: Wow, great catch! Somehow this order happened, I am a little upset that this slipped by me.

gengokit/httptransport/templates/server.go
adamryman: This is a raw string literal. Using the back quote or grave symbol `. We do this so we do not have to escape quotes. Though the only way to have a back quote in a raw string literal is to concatenate it with a string using quotes.

> [Raw string literals are character sequences between back quotes, as in `. Within the quotes, any character may appear except back quote.](https://golang.org/ref/spec#String_literals)

hasIan: I'm familiar with raw strings but have never seen this. The context missing was the open backtick at the very top of the template and the one at the end of the template.

PR #191, opened 2017-06-13
[MVS-119] Update sever error encoder to set status code and headers.

adamryman:
- [x] Update errorEncoder to type cast against `Headerer` and `StatusCoder` for custom http status codes.
    - [x] Add `type httpError struct` somewhere in truss
    - [x] Test that HTTP status code is set on response
    - [x] Test that HTTP Headers are set on response
    - [x] Test that gRPC transport is unaffected by error with StatusCode
    - [x] Update improper content type and/or malformed json to return StatusCode `http.StatusBadRequest` `401` [BLOCKED by #190, will be editing same area of code]


cmd/_integration-tests/transport/http_test.go
```diff
@@ -400,6 +400,67 @@ func TestStrangeRPCName(t *testing.T) {
 	}
 }
 
+func TestHTTPErrorStatusCodeAndNilHeaders(t *testing.T) {
+	// See handlers/handlers.go for implementation
+	// Returns status code http.StatusTeapot
+	req, err := http.NewRequest("GET", httpAddr+"/status/code", strings.NewReader(""))
```
hasLeland: Like I mentioned in the other PR, it's possible to make a GET request with a body, but it's not usual and it seems strange to do it in a test. Making these POST requests would be clearer.

hasLeland (approved):
Hooray, we get errors now!

cmd/_integration-tests/transport/http_test.go
adamryman: There is no body? It is empty.  Do you mean a nil? That can happen.

cmd/_integration-tests/transport/http_test.go
```diff
@@ -400,6 +400,67 @@ func TestStrangeRPCName(t *testing.T) {
...
+	defer httpResp.Body.Close()
+
+	got, want := httpResp.Header, map[string][]string{
+		"Foo":  []string{"Bar"},
+		"Test": []string{"A", "B"},
+	}
+
+	for k := range want {
+		if len(got[k]) == 0 {
+			t.Fatalf("Expected Header `%s:%s`; Got not header with key: `%s`", k, want[k], k)
```
hasIan: Typo: "Got not"

This is a controlled environment, so this seems okay, but testing that the length is zero is not indicative of a lack of map membership--for this to be certain, you must use the two-return-value map access, `v, ok := m[k]`.

Curious about the backtick quotes in the error strings. NBD, but seems kind of arbitrary since backtick quotes are not typically used in general.

gengokit/httptransport/templates/server.go
```diff
@@ -179,12 +179,33 @@ func MakeHTTPHandler(ctx context.Context, endpoints Endpoints, logger log.Logger
...
 
+// ErrorEncoder writes the error to the ResponseWriter, by default a content
+// type of application/json, a body of json with key "error" and the value
+// error.Error(), and a status code of 500. If the error implements Headerer,
+// the provided headers will be applied to the response. If the error
+// implements json.Marshaler, and the marshaling succeeds, the JSON encoded
+// form of the error will be used. If the error implements StatusCoder, the
+// provided StatusCode will be used instead of 500.
 func errorEncoder(_ context.Context, err error, w http.ResponseWriter) {
+	contentType := "application/json; charset=utf-8"
```
hasIan: This should probably be a constant. Although it's only used once so someone could probably make the argument that it should just be used inline, but I don't feel strongly either way here.

hasIan (approved):
Couple comments, but looks nice.

cmd/_integration-tests/transport/http_test.go
hasLeland: Yes, wow I was literally imagining there was some body there; but there isn't :sheep: but yeah, those bodies can be `nil` and it would be a tad clearer

truss/httperr.go
```diff
@@ -0,0 +1,28 @@
+package truss
+
+import "net/http"
+
+// HTTPError encodes information in an error type for the
+// github.com/go-kit/kit/tranport/http.DefaultErrorEncoder to create a
+// meaningful http response
+func HTTPError(err error, statusCode int, headers http.Header) error {
```
hasLeland: Since we don't want `truss` to be a lib, I think this HTTPError type and struct should be moved to generated code. I feel it makes sense for it to live in `name-service/svc/transport_http.go` or potentially in a new file or sub-package within `name-service/svc/`.

hasLeland (changes requested):

truss/httperr.go
adamryman: It is now in `transport_http.go` and is unexported. As promoting importing a service inside a lib seems undesired. If a lib wants to use the `StatusCoder` and `Headerer` functionality then they can copy and paste this into their implementation.

hasLeland (approved):

PR #194, opened 2017-06-23
Expand function type in WrapAllLabeledExcept

adamryman:
This allows an interface such as:
```
type Wrapper interface {
    WrapAllLabeledExcept(func(string, endpoint.Endpoint)
    endpoint.Endpoint, exclude ...string)
}
```
to compile. Before you would have to import the service and use the
`LabeledMiddleware` type.

Go 1.9 could allows us to not need this, though that is low priority.


zaquestion (approved):

zaquestion:
approved pending tests

PR #195, opened 2017-06-26
Add default flag and protobuf file as output

hasLeland:
This PR adds the `--default` flag to truss, which gives you a basic service-compatible protobuf file "out of the box". However, there's several items I want to bring to the attention of reviewers:

1. Do we want this feature to only output a `.proto` file, or should it also generate a service? If so, how are names for services in the generated proto file chosen? Where should the service be located?
2. Currently, the default proto file which is created has a single RPC, called `Status`, based on what we use within multiverse. Are there other RPC's to include in the default proto file?
3. The actual name of the package and service in this default file cannot be named `default`, as that's a reserved word in the protocol buffer language. To get this up fast, I just removed the vowels and used `Dflt`, but that's pretty gross. So, we may want to chose a different name, such as "example", "initial", or "sample". Feel free to suggest others.

### DIFF
--- a/cmd/truss/main.go
+++ b/cmd/truss/main.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/TuneLab/go-truss/truss"
 	"github.com/TuneLab/go-truss/truss/execprotoc"
+	"github.com/TuneLab/go-truss/truss/getstarted"
 	"github.com/TuneLab/go-truss/truss/parsesvcname"
 
 	"github.com/TuneLab/go-truss/deftree"
@@ -29,6 +30,7 @@ var (
 	svcPackageFlag = flag.String("svcout", "", "Go package path where the generated Go service will be written. Trailing slash will create a NAME-service directory")
 	verboseFlag    = flag.BoolP("verbose", "v", false, "Verbose output")
 	helpFlag       = flag.BoolP("help", "h", false, "Print usage")
+	getStartedFlag = flag.BoolP("getstarted", "", false, "Output a 'getstarted.proto' protobuf file in ./")
 )
 
 var binName = filepath.Base(os.Args[0])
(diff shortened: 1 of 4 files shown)


cmd/truss/main.go
```diff
@@ -29,6 +30,7 @@ var (
 	svcPackageFlag = flag.String("svcout", "", "Go package path where the generated Go service will be written. Trailing slash will create a NAME-service directory")
 	verboseFlag    = flag.BoolP("verbose", "v", false, "Verbose output")
 	helpFlag       = flag.BoolP("help", "h", false, "Print usage")
+	defaultFlag    = flag.BoolP("default", "", false, "Output a default protobuf file in ./")
```
hasAdamr: Maybe mention the file name the it will create here?

truss/defaultsvc/bindata.go
```diff
@@ -0,0 +1,235 @@
+// Code generated by go-bindata.
```
hasAdamr: Could generating this be added to the root Makefile?

truss/defaultsvc/default.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto3";
+
+package dflt;
```
hasAdamr: Why not `default`?

truss/defaultsvc/default.proto
```diff
@@ -0,0 +1,26 @@
...
+ 
+import "github.com/TuneLab/go-truss/deftree/googlethirdparty/annotations.proto";
+
+service Dflt {
```
hasAdamr: Why not `default`?

hasAdamr (changes requested):
Looks great! 

I would like the generation of the file added to the `Makefile`. Or at least a script somewhere. I know it is just a `gobindata` command. Though that is not something I want to remember.

hasAdamr:
Idea: It could be possible to pass parameters to this flag. Basically just a service name that gets templated into the package name and service name.

PR #196, opened 2017-06-28
Add --getstarted flag for creating a protobuf file

hasLeland:
This PR adds the `--getstarted` flag to truss, which gives you a basic service-compatible protobuf file "out of the box". Here's what its output looks like:

```
$ truss --getstarted
INFO[0000] A "starter" protobuf file named 'getstarted.proto' has been created in the
current directory. You can generate a service based on this new protobuf file
at any time using the following command:

    truss getstarted.proto

If you want to generate a protofile with a different name, use the
'--getstarted' option with the name of your choice after '--getstarted'. For
example, to generate a 'foo.proto', use the following command:

    truss --getstarted foo
 
```

## Notes for reviewers:

In addition to the usual code review, please edit the english help messages in the `truss/getstarted/template.go` for consistency, tone, and grammar.

### DIFF
--- a/cmd/truss/main.go
+++ b/cmd/truss/main.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/TuneLab/go-truss/truss"
 	"github.com/TuneLab/go-truss/truss/execprotoc"
+	"github.com/TuneLab/go-truss/truss/getstarted"
 	"github.com/TuneLab/go-truss/truss/parsesvcname"
 
 	"github.com/TuneLab/go-truss/deftree"
@@ -29,6 +30,7 @@ var (
 	svcPackageFlag = flag.String("svcout", "", "Go package path where the generated Go service will be written. Trailing slash will create a NAME-service directory")
 	verboseFlag    = flag.BoolP("verbose", "v", false, "Verbose output")
 	helpFlag       = flag.BoolP("help", "h", false, "Print usage")
+	getStartedFlag = flag.BoolP("getstarted", "", false, "Output a 'getstarted.proto' protobuf file in ./")
 )
 
 var binName = filepath.Base(os.Args[0])
@@ -77,15 +79,23 @@ func main() {
 		os.Exit(0)
 	}
 
+	log.SetLevel(log.InfoLevel)
+	if *verboseFlag {
+		log.SetLevel(log.DebugLevel)
+	}
+
+	if *getStartedFlag {
+		pkg := ""
+		if len(flag.Args()) > 0 {
+			pkg = flag.Args()[0]
+		}
+		os.Exit(getstarted.Do(pkg))
+	}
+
 	if len(flag.Args()) == 0 {
 		fmt.Fprintf(os.Stderr, "%s: missing .proto file(s)\n", binName)
 		flag.Usage()
 		os.Exit(1)
-	}
-
-	log.SetLevel(log.InfoLevel)
-	if *verboseFlag {
-		log.SetLevel(log.DebugLevel)
 	}
 
 	cfg, err := parseInput()
--- a/truss/getstarted/getstarted.go
+++ b/truss/getstarted/getstarted.go
@@ -1,0 +1,121 @@
+package getstarted
+
+import (
+	"bytes"
+	"os"
+	"strings"
+	"text/template"
+
+	log "github.com/Sirupsen/logrus"
+	gogen "github.com/golang/protobuf/protoc-gen-go/generator"
+	"github.com/pkg/errors"
+)
+
+type protoInfo struct {
+	alias string
+}
+
+func (p protoInfo) FileName() string {
+	return p.PackageName() + ".proto"
+}
+
+func (p protoInfo) PackageName() string {
+	a := p.alias
+	a = strings.Replace(a, "-", "", -1)
+	a = strings.Replace(a, " ", "", -1)
+
+	a = strings.ToLower(a)
+	return a
+}
+
+func (p protoInfo) ServiceName() string {
+	a := p.alias
+	a = strings.Replace(a, "-", "_", -1)
+	a = strings.Replace(a, " ", "_", -1)
+	return gogen.CamelCase(a)
+}
+
+// Do writes a default protobuf file to the current directory, in a file named
+// based on the 'pkg' param, defaulting to "getstarted.proto" if pkg is empty.
+// If the file exists, Do prints a warning and returns a non-zero exit code.
+// The non-zero exit code is to enable using the return from this function in
+// os.Exit().
+func Do(pkg string) int {
+	const fallbackFName = "get_started"
+	if pkg == "" {
+		pkg = fallbackFName
+	}
+	pkg = removeDotProtoSuffix(pkg)
+	pinfo := protoInfo{
+		alias: pkg,
+	}
+	if _, err := os.Stat(pinfo.FileName()); err == nil {
+		existingFile, err := renderTemplate("existingFileMsg", existingFileMsg, pinfo)
+		if err != nil {
+			log.Error(err)
+			return 1
+		}
+		log.Error(string(existingFile))
+		return 1
+	}
+	f, err := os.Create(pinfo.FileName())
+	if err != nil {
+		log.Error(errors.Wrapf(err, "cannot create %q", pinfo.FileName()))
+		return 1
+	}
+
+	code, err := renderTemplate(pinfo.FileName(), starterProto, pinfo)
+	if err != nil {
+		log.Error(err)
+		return 1
+	}
+
+	_, err = f.Write(code)
+	if err != nil {
+		log.Error(errors.Wrapf(err, "cannot write default contents to %q", pinfo.FileName()))
+		return 1
+	}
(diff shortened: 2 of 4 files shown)


truss/getstarted/getstarted.go
```diff
@@ -0,0 +1,116 @@
...
+
+// removeDotProto exists to preempt and warn a user who enters a name
+// containing `.proto`. It will warn the user of their incorrect input and will
+// demonstrate how their input can be corrected. Then, the program continues
+// using the corrected input it warned about.
+func removeDotProto(pkg string) string {
+	want := strings.Replace(pkg, ".proto", "", -1)
+	if strings.HasSuffix(pkg, ".proto") {
+		executor := struct{ Got, Want string }{pkg, want}
+		warn, err := renderTemplate("dotProtoInName", dotProtoInName, executor)
```
hasAdamr: I feel like the `dotProtoInName` template could be an inline constant in this function. 

I don't think it will be used anywhere else. And it is nice to read the message in context.

truss/getstarted/getstarted.go
```diff
@@ -0,0 +1,116 @@
...
+// Do writes a default protobuf file to the current directory, in a file named
+// "default.proto". If the file exists, it prints a warning and returns a
+// non-zero exit code. The non-zero exit code is to enable using the return
+// from this function in os.Exit().
+func Do(pkg string) int {
+	const fallbackFName = "get_started"
+	if pkg == "" {
+		pkg = fallbackFName
+	}
+	pkg = removeDotProto(pkg)
```
hasAdamr: Nit: To me, this function sounds like it will remove a `.proto` file.

truss/getstarted/getstarted.go
```diff
@@ -0,0 +1,116 @@
...
+	pinfo := protoInfo{
+		alias: pkg,
+	}
+	// For convenience sake, render the help messages now. Saves us from doing
+	// nested error checks later, making the code a bit clearer.
+	nextStep, err := renderTemplate("nextStepMsg", nextStepMsg, pinfo)
```
hasAdamr: Possibly move rendering `nextStep` to just before you use it. Kind of out of place to have it here from my perspective.

truss/getstarted/getstarted.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (p protoInfo) FileName() string {
+	return p.PackageName() + ".proto"
+}
+
+func (p protoInfo) PackageName() string {
+	a := p.alias
+	a = strings.Replace(a, "-", "", -1)
+	a = strings.Replace(a, "_", "", -1)
```
hasAdamr: Are package names not allowed to have `_`'s?

truss/getstarted/getstarted.go
```diff
@@ -0,0 +1,116 @@
...
+	return want
+}
+
+func renderTemplate(name string, tmpl string, executor interface{}) ([]byte, error) {
+	codeTemplate := template.Must(template.New(name).Parse(tmpl))
+
+	code := bytes.NewBuffer(nil)
+	err := codeTemplate.Execute(code, executor)
+	if err != nil {
+		return []byte{}, errors.Wrapf(err, "attempting to execute template %q", name)
```
hasAdamr: Maybe:
```
return nil, errors.Wrap(...
```

truss/getstarted/getstarted.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (p protoInfo) ServiceName() string {
+	a := p.alias
+	a = strings.Replace(a, "-", "_", -1)
+	a = strings.Replace(a, " ", "_", -1)
+	return gogen.CamelCase(a)
+}
+
+// Do writes a default protobuf file to the current directory, in a file named
+// "default.proto". If the file exists, it prints a warning and returns a
```
hasAdamr: `default.proto`?

hasAdamr (commented):
Looking good! 

A few things I would like to hear your reasoning for and/or changed before I click the approve button.

Will be nice to have this for creating a new service. :open_hands:

truss/getstarted/getstarted.go
hasLeland: There's nothing preventing them from having underscores; when implementing this I removed underscores due to some notion of aesthetics that was in my brain. I'll remove that if you think they should be allowed.

truss/getstarted/getstarted.go
hasLeland: Whoops, should be changed to `getstarted.proto`; good catch.

truss/getstarted/getstarted.go
hasLeland: Good point, changing to `removeDotProtoSuffix`

truss/getstarted/getstarted.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (p protoInfo) ServiceName() string {
+	a := p.alias
+	a = strings.Replace(a, "-", "_", -1)
+	a = strings.Replace(a, " ", "_", -1)
+	return gogen.CamelCase(a)
+}
+
+// Do writes a default protobuf file to the current directory, in a file named
+// "getstarted.proto". If the file exists, it prints a warning and returns a
```
hasAdamr: Won't the file name be based on what is passed to this function?

truss/getstarted/getstarted.go
```diff
@@ -0,0 +1,121 @@
...
+func Do(pkg string) int {
+	const fallbackFName = "get_started"
+	if pkg == "" {
+		pkg = fallbackFName
+	}
+	pkg = removeDotProtoSuffix(pkg)
+	pinfo := protoInfo{
+		alias: pkg,
+	}
+	existingFile, err := renderTemplate("existingFileMsg", existingFileMsg, pinfo)
```
hasAdamr: What happened to the comment about this being rendered early? It is kind of confusing by itself.

truss/getstarted/getstarted.go
```diff
@@ -0,0 +1,121 @@
...
+
+	code, err := renderTemplate(pinfo.FileName(), starterProto, pinfo)
+	if err != nil {
+		log.Error(err)
+		return 1
+	}
+
+	_, err = f.Write(code)
+	if err != nil {
+		log.Error(errors.Wrapf(err, "cannot write default contents to %q", pinfo))
```
hasAdamr: `pinfo.FileName()` correct?

cmd/truss/main.go
```diff
@@ -29,6 +30,7 @@ var (
 	svcPackageFlag = flag.String("svcout", "", "Go package path where the generated Go service will be written. Trailing slash will create a NAME-service directory")
 	verboseFlag    = flag.BoolP("verbose", "v", false, "Verbose output")
 	helpFlag       = flag.BoolP("help", "h", false, "Print usage")
+	getStartedFlag = flag.BoolP("getstarted", "", false, "Output a 'getstarted.proto' protobuf file in ./")
```
hasAdamr, line 33: Should there be some documentation about how you can pass something to this to get your own name for the file/package/servicename?

truss/getstarted/getstarted.go
hasAdamr: Kind of odd to me. Really I don't mind either way. So I guess leave it? ¯\\\_(ツ)_/¯

hasAdamr (commented):
Couple more things before that Approve checkmark.

Thanks for looking over the feedback.

cmd/truss/main.go
hasLeland: There is documentation, shown when you use the option. https://github.com/TuneLab/go-truss/pull/196/files#diff-b7a48309a1d9f912b7f8857bd593df6eR38

Because the `getStartedFlag` is actually a bool flag and the name you can provide is actually handled outside of the way the flag package canonically does assignment, I don't want to say in the help message that you can provide a name. If we did, someone might think it means you can pass `--getstarted=foo` which will not work and will cause the program to fail.

So instead, I put the longer, more explicit instructions on how to do it in the `nextSteps` message printed when you actually use `--getstarted`.

hasAdamr (approved):
Great! :+1: 

Though wait for a review from either @zaquestion or @hasIan before merging though. Maybe ping them if you want to get this in sooner rather than later.

truss/getstarted/getstarted.go
zaquestion: If its a legal in protobuf/go it should probably be allowed and not enforced by us.

hasLeland: Alright, taking it out.

truss/getstarted/template.go
```diff
@@ -0,0 +1,48 @@
...
+syntax = "proto3";
+
+package {{.PackageName}};
+ 
+import "github.com/TuneLab/go-truss/deftree/googlethirdparty/annotations.proto";
+
+service {{.ServiceName}} {
+  rpc Status(StatusRequest) returns (StatusResponse) {
+    option (google.api.http) = {
+      get: "/Status"
```
zaquestion: Thoughts on lower casing this? Believe that would be more standard for REST

hasLeland: It was that way originally because it was copy-pasted from a multiverse interface definition, but having it be lowercase is an improvement. Will make the change.

PR #197, opened 2017-06-29
Remove trailing whitespace from getstarted template

adamryman:


truss/getstarted/template.go
```diff
@@ -4,7 +4,7 @@ const starterProto = `
 syntax = "proto3";
 
 package {{.PackageName}};
-
```
hasLeland, line 7: :anguished: :+1:

hasLeland (approved):

PR #199, opened 2017-07-12
Initial work: Add benchmarks for truss HTTP pipeline

hasLeland:
This PR adds benchmarks for the HTTP pipeline of truss-generated services. Additionally, this PR adds those same benchmarks to the build pipeline so they'll always be run.

Please let me know if there are other areas or things which should be benchmarked, or if the way that I'm currently benchmarking things is incorrect.

-----

To run these benchmarks yourself, run the following command:

    cd $GOPATH/src/github.com/TuneLab/truss/cmd/_integration-tests/transport && make

I'm pasting the results of the benchmark here for posterities sake:
```
go test -run=XXX -bench=.
BenchmarkGetWithQueryClient-4             	   10000	    138193 ns/op
BenchmarkGetWithQueryClient_NoNetwork-4   	  200000	      8018 ns/op
BenchmarkClientEncoding-4                 	  300000	      4092 ns/op
BenchmarkAddition-4                       	20000000	        64.9 ns/op
PASS
```

### DIFF
--- a/cmd/_integration-tests/transport/http_benchmarks_test.go
+++ b/cmd/_integration-tests/transport/http_benchmarks_test.go
@@ -1,11 +1,17 @@
 package test
 
 import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"math/rand"
 	"net/http"
 	"net/http/httptest"
+	"os"
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	// 3d Party
 	"golang.org/x/net/context"
@@ -24,9 +30,7 @@ var assignval interface{}
 
 // See how fast it is to make a full RPC complete with actual (local) HTTP connection
 func BenchmarkGetWithQueryClient(b *testing.B) {
-	var req pb.GetWithQueryRequest
-	req.A = 12
-	req.B = 45360
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
 
 	svchttp, err := httpclient.New(httpAddr, httpclient.CtxValuesToSend("request-url", "transport"))
 	if err != nil {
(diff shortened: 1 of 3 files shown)


zaquestion:
PR'd this branch with some ideas, https://github.com/hasLeland/truss/pull/3

cmd/_integration-tests/transport/http_benchmarks_test.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+type ServerTestData struct {
+	r        *http.Request
+	recorder *httptest.ResponseRecorder
+}
+
+// Instead of actually sending a request over the network, instead we construct
+// the service endpoints within this function and call it's method directly.
+func BenchmarkGetWithQueryClient_NoNetwork(b *testing.B) {
```
hasAdamr, line 60: It should be noted that this tests the servers http decode and http decode functions, endpoints, and the actual handler. Though it does not test the speed of the client encode or client decode.

cmd/_integration-tests/transport/http_benchmarks_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/TuneLab/truss/cmd/_integration-tests/transport/transportpermutations-service/svc"
+
+	httptransport "github.com/go-kit/kit/transport/http"
+)
+
+var assignval interface{}
+
+// See how fast it is to make a full RPC complete with actual (local) HTTP connection
+func BenchmarkGetWithQueryClient(b *testing.B) {
+	var req pb.GetWithQueryRequest
```
hasAdamr: I feel we should use and Endpoint that uses query params, body params, and path params. Just so our benchmark includes all of these different features of http transport.

cmd/_integration-tests/transport/http_benchmarks_test.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Benchmark the speed of encoding an HTTP request
+func BenchmarkClientEncoding(b *testing.B) {
+	var req pb.GetWithQueryRequest
+	req.A = 12
+	req.B = 45360
+	for i := 0; i < b.N; i++ {
+		httpreq, _ := http.NewRequest("GET", "http://localhost/getwithquery", strings.NewReader(string("")))
+		httpclient.EncodeHTTPGetWithQueryZeroRequest(context.Background(), httpreq, &req)
```
hasAdamr, line 114: Again I believe we should use an endpoint that has query, path, and body parameters so we test all the features of our http transport.

zaquestion:
retroactive LGTM

PR #200, opened 2017-07-17
Add Windows installation script

pauln:
Although Truss will happily work under Windows, installing it is not a trivial process.  I've written a batch file which installs Truss by following the process which runs when calling `make dependencies; make`, and so far (tested on a couple of different Windows PCs) seems to reliably build and install Truss.

The main deviation from the Makefile-driven build process is that Windows has no native way to convert a Unix timestamp into the format used for the version date - so I wrote a super-quick, super-basic date conversion tool in Go ([pauln/go-datefmt](https://github.com/pauln/go-datefmt)) which the batch file installs and uses.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,12 @@ Otherwise [install from source.](https://github.com/google/protobuf)
 	make dependencies
 	make
 	```
+	On Windows, do the following instead:
+	```
+	go get -u -d github.com/TuneLab/truss
+	cd %GOPATH%/src/github.com/TuneLab/truss
+	wininstall.bat
+	```
 
 ## Usage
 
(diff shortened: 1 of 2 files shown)


hasIan:
Thank you for the contribution, Paul! Admittedly, Truss was developed on macOS and Linux, serving our internal needs primarily. We'd be thrilled for it to be more usable for others.

I'm missing where the installation batch file installs your `go-datefmt` tool, though: I only see where it is invoked. Is this an oversight, or do I have Monday Brain?

pauln:
Ah, nuts - I must have missed that! Naturally, it was installed for me since I built it - and my colleague who tested the install script also had it installed, as I'd asked him to test installing/using go-datefmt!

hasIan:
Looks good now. :shipit: 🎉 

Welcome, Windows Truss users. 😄

PR #204, opened 2017-07-24
Move middlewares into handlers package

adamryman:
In order to configure middlewares using the `NewService` function in a more straight forward way, the `middlewares` package has been moved into `handlers`. 

This means that the function `WrapEndpoints` is now in the same package as `NewService`, allowing the middlewares to be configured with package level variables rather than cross package variables.

This is not a complete solution, but rather an incremental improvement to the current configuration situation.

There was no real reason that `middlewares` became its own package, just a decision I made along the line. Both @zaquestion and @hasLeland agree that it makes sense to be part of the handlers package.

To help with the transition, running these commands in the `NAME-service` directory in addition to rerunning truss should result in a all old middlewares still functioning.

```
cat middlewares/endpoints.go | \
    sed 's|^package middlewares|package handlers|' > handlers/middlewares.go

# Note that this creates a file `service_middlewares.go` that truss will not create, 
# though this is easier than trying to combine the files together here.

cat middlewares/service.go | \
    sed 's|^package middlewares|package handlers|' > handlers/service_middlewares.go

rm middlewares/endpoints.go
rm middlewares/service.go
rmdir middlewares

```

Additionally:
- rename package `github.com/TuneLab/truss/gengokit/handler` to `github.com/TuneLab/truss/gengokit/handlers`
- add `NewEndpoints` to package `svc/server`


gengokit/middlewares/middlewares.go
```diff
@@ -1,63 +0,0 @@
...
-		if m.prevEndpoints != nil {
-			return m.prevEndpoints, nil
-		}
-		return data.ApplyTemplate(templates.EndpointsBase, "Endpoint")
-	case ServicePath:
-		if m.prevService != nil {
-			return m.prevService, nil
-		}
-		return data.ApplyTemplate(templates.ServiceBase, "Service")
-	}
```
hasLeland, line 61: I love that this logic is collapsed in the new version

hasLeland (approved):
In general, a very nice little cleanup of code!

zaquestion (approved):

Issue #202, opened 2017-07-24
Comments between package name and imports cause `panic: New node is nil`


pauln:
This one had us scratching our heads for a while!

If you have any comments between the `package` line and the first `import`, Truss fails with the following error:
```
panic: New node is nil, namepath: '[echo.proto github.com/TuneLab/truss/deftree/googlethirdparty/annotations.proto]' cur_node: '&{<nil> echo.proto  [0xc04239d270 0xc04239d2c0 0xc04239d310] [] [0xc042034480]}'


goroutine 1 [running]:
github.com/TuneLab/truss/deftree.(*MicroserviceDefinition).SetComment(0xc0422dfcc0, 0xc04224f380, 0x2, 0x2, 0xc04235fbe8, 0x4)
        C:/Users/Paul/go/src/github.com/TuneLab/truss/deftree/deftree.go:151 +0x1f7
github.com/TuneLab/truss/deftree.AssociateComments(0x8c7260, 0xc0422dfcc0, 0xc0422de190)
        C:/Users/Paul/go/src/github.com/TuneLab/truss/deftree/associate_comments.go:269 +0x944
github.com/TuneLab/truss/deftree.New(0xc0422de190, 0x8c1e60, 0xc042004720, 0xc042004720, 0x0, 0x0, 0xc0422de190)
        C:/Users/Paul/go/src/github.com/TuneLab/truss/deftree/build_deftree.go:72 +0x509
main.parseServiceDefinition(0xc042050780, 0x0, 0x744b67, 0x12, 0x0, 0x0)
        C:/Users/Paul/go/src/github.com/TuneLab/truss/cmd/truss/main.go:283 +0x5a7
main.main()
        C:/Users/Paul/go/src/github.com/TuneLab/truss/cmd/truss/main.go:104 +0x126
```

Removing the offending comment(s) allows Truss to operate correctly.

An example of troublesome comments:
```
syntax = "proto3";
package echo;

// This comment will cause Truss to panic.
import "github.com/TuneLab/truss/deftree/googlethirdparty/annotations.proto";
```

We encountered this issue as we had comments separating our imports into `// External imports` and `// Internal imports`, but it seems that any comment at all before the first import will trigger this issue.

hasLeland:
@pauln thank you for the bug report! I can't give an exact timeline on this bug, but we've added it to the team sprint and I'll be digging into fixing this.

[...]

PR #206, opened 2017-07-24
Fix bug reported in issue #202, comment causing panics in documentation generation

hasLeland:
This PR implements a fix for the problem reported by @pauln in issue #202. The fix implemented here is the one I roughly outlined in a comment on that issue:

> ...it seems that `truss` is written with a bug in it such that if you attempt to associate a comment with something that this portion of the codebase doesn't care about or keep track of, then truss incorrectly panics instead of returning an error which can be handled.
> 
> IMO, the fix is to change [this function](https://github.com/TuneLab/truss/blob/096ba69249ea9c36208500413b665307c7431d00/deftree/deftree.go#L151) so that it returns an error which can be logged/handled/ignored as needed.

More specifically, I've changed that function to now return custom-typed errors, which are type-asserted by the consuming function to check whether they are of this type which can optionally be handled, and if they they are, then an error is logged.

Because this portion of the code (`deftree/`) is responsible for documentation generation, which is lower priority than code generation, I opted to only log an error instead of terminating program execution.

### DIFF
--- a/deftree/deftree.go
+++ b/deftree/deftree.go
@@ -72,7 +72,7 @@ type Describable interface {
 // documentation on these Methods.
 type Deftree interface {
 	Describable
-	SetComment([]string, string)
+	SetComment([]string, string) error
 	String() string
 }
 
@@ -142,17 +142,20 @@ func (self *MicroserviceDefinition) GetByName(name string) Describable {
 // on the result of the last call, beginning with this MicroserviceDefinition.
 // Once the final Describable object is found, the `description` field of that
 // struct is set to `comment_body`.
-func (self *MicroserviceDefinition) SetComment(namepath []string, comment_body string) {
+//
+// If a node cannot be found with the provided namepath, returns an error.
+func (self *MicroserviceDefinition) SetComment(namepath []string, comment_body string) error {
 	var cur_node Describable
 	cur_node = self
 	for _, name := range namepath {
 		new_node := cur_node.GetByName(name)
 		if new_node == nil {
-			panic(fmt.Sprintf("New node is nil, namepath: '%v' cur_node: '%v'\n", namepath, cur_node))
+			return fmt.Errorf("cannot find node with name %q in namepath %q on cur_node %q", name, namepath, cur_node)
 		}
 		cur_node = new_node
 	}
 	cur_node.SetDescription(comment_body)
+	return nil
 }
 
 // String kicks off the recursive call to `describe` within the tree of
--- a/deftree/associate_comments.go
+++ b/deftree/associate_comments.go
@@ -266,7 +266,10 @@ func AssociateComments(dt Deftree, req *plugin.CodeGeneratorRequest) {
 					if err != nil {
 						log.Debugf("Couldn't place comment '%v' due to error traversing tree: %v", cleanStr(lead), err)
 					} else {
-						dt.SetComment(name_path, scrubComments(lead))
+						err = dt.SetComment(name_path, scrubComments(lead))
+						if err != nil && !strings.Contains(err.Error(), "cannot find node") {
+							log.Errorf("cannot set comment: %v", err)
+						}
 					}
 				}
 			}
--- a/deftree/associate_comments_test.go
+++ b/deftree/associate_comments_test.go
@@ -46,3 +46,43 @@ func TestCommentedEnum(t *testing.T) {
 		t.Fatalf("Comment found in Enum is not expected; got = %q, want = %q", got, want)
 	}
 }
+
+// Test to ensure that placing comments directly above a proto3 import functions correctly.
+func TestCommentedImport(t *testing.T) {
+	// Create our request, then assemble a basic deftree
+	defStr := `
+		// This comment should not cause any problems
+		syntax = "proto3";
+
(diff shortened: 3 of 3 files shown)


deftree/deftree.go
```diff
@@ -136,23 +136,40 @@ func (self *MicroserviceDefinition) GetByName(name string) Describable {
 	return nil
 }
 
+type optionErr struct {
```
zaquestion: IIRC, we managed to remove a lot of the optional error usages throughout the repo in previous changes. Personally I'd rather define a error in this package and check the Type in the consuming function over incorporating the optional error again.

zaquestion (approved):

deftree/deftree.go
```diff
@@ -136,23 +136,36 @@ func (self *MicroserviceDefinition) GetByName(name string) Describable {
...
+// If a node cannot be found with the provided namepath, returns an error.
+func (self *MicroserviceDefinition) SetComment(namepath []string, comment_body string) error {
 	var cur_node Describable
 	cur_node = self
 	for _, name := range namepath {
 		new_node := cur_node.GetByName(name)
 		if new_node == nil {
-			panic(fmt.Sprintf("New node is nil, namepath: '%v' cur_node: '%v'\n", namepath, cur_node))
+			return notFoundErr{
+				err: fmt.Errorf("cannot find node with name %q in namepath %q on cur_node %q\n", name, namepath, cur_node),
```
hasIan: You don't need `\n` in an error message.

deftree/deftree.go
```diff
@@ -136,23 +136,36 @@ func (self *MicroserviceDefinition) GetByName(name string) Describable {
 	return nil
 }
 
+type notFoundErr struct {
```
hasIan: This seems like it doesn't need to be a struct.
```
type notFoundErr error
```
would be sufficient, and it can be instantiated like
```
err := notFoundErr(fmt.Errorf("..."))
```

hasIan (approved):
Looks good; just a couple comments for minor simplification.

deftree/deftree.go
hasLeland: So, in testing this out I found that doing this does not let us discern between different errors. For example, see the following file:
```
// Demonstration that these all appear as the all types according to type
// assertions, as they all fulfill interfaces which are equivalent. So there's
// no way to differentiate.
package main

import "fmt"

type thing struct {
	err error
}

func (t thing) Error() string {
	return t.err.Error()
}

type secondThing error

func main() {
	x := func() error {
		return thing{fmt.Errorf("test")}
	}()
	y := func() error {
		return secondThing(fmt.Errorf("test"))
	}()
	_, ok1 := x.(error)
	fmt.Printf("x is error: %v\n", ok1)
	_, ok2 := x.(thing)
	fmt.Printf("x is thing: %v\n", ok2)
	_, ok3 := x.(secondThing)
	fmt.Printf("x is secondThing: %v\n", ok3)

	_, ok1 = y.(error)
	fmt.Printf("y is error: %v\n", ok1)
	_, ok2 = y.(secondThing)
	fmt.Printf("y is secondThing: %v\n", ok2)
	_, ok3 = y.(secondThing)
	fmt.Printf("y is secondThing: %v\n", ok3)
}
```
If I run this, you can see that all variables appear to be all types at the same time:

```$ go run test.go 
x is error: true
x is thing: true
x is secondThing: true
y is error: true
y is secondThing: true
y is secondThing: true
```

hasLeland: @zaquestion  @hasIan In testing this out I found that doing this does not let us discern between different errors. For example, see the following file:
```
// Demonstration that these all appear as the all types according to type
// assertions, as they all fulfill interfaces which are equivalent. So there's
// no way to differentiate.
package main

import "fmt"

type thing struct {
	err error
}

func (t thing) Error() string {
	return t.err.Error()
}

type secondThing error

func main() {
	x := func() error {
		return thing{fmt.Errorf("test")}
	}()
	y := func() error {
		return secondThing(fmt.Errorf("test"))
	}()
	_, ok1 := x.(error)
	fmt.Printf("x is error: %v\n", ok1)
	_, ok2 := x.(thing)
	fmt.Printf("x is thing: %v\n", ok2)
	_, ok3 := x.(secondThing)
	fmt.Printf("x is secondThing: %v\n", ok3)

	_, ok1 = y.(error)
	fmt.Printf("y is error: %v\n", ok1)
	_, ok2 = y.(secondThing)
	fmt.Printf("y is secondThing: %v\n", ok2)
	_, ok3 = y.(secondThing)
	fmt.Printf("y is secondThing: %v\n", ok3)
}
```
If I run this, you can see that all variables appear to be all types at the same time:

```$ go run test.go 
x is error: true
x is thing: true
x is secondThing: true
y is error: true
y is secondThing: true
y is secondThing: true
```
-----
As far as I can tell, the only options remaining which I know do work are:
1. Go back to optional interfaces
2. Use string comparisons

hasLeland: I opted to go with string comparisons for brevity.

hasLeland:
It should be noted that the ultimate solution for this, though it lacks online discussion of the particular changes made, was discussed offline with @zaquestion

PR #210, opened 2017-07-27
Move to Gorilla Mux

### LINKED ISSUE #209: Generated HTTP transport can't cope with different HTTP verbs on the same URI
The HTTP transport generated by Truss doesn't appear to handle using the same URI with different HTTP verbs (i.e. GET vs POST) for different RPC methods.  This means that it's not able to generate a clean RESTful API, as you'd normally have something along the lines of:
```proto
service Foobar {
  rpc GetFoos (GetFoosRequest) returns (GetFoosResponse) {
    option (google.api.http) = {
      get: "/foo"
    };
  }

  rpc CreateFoo (CreateFooRequest) returns (CreateFooResponse) {
    option (google.api.http) = {
      post: "/foo"
      body: "*"
    };
  }
}
```

However, this results in the following error:
```
panic: http: multiple registrations for /foo

goroutine 23 [running]:
net/http.(*ServeMux).Handle(0xc04216e1e0, 0x99d904, 0x14, 0xbf60a0, 0xc042168200)
	C:/Go/src/net/http/server.go:2254 +0x617
foobar/foobar-service/svc.MakeHTTPHandler(0xf8c438, 0xc0420384a0, 0xc042137c40, 0xc042137c60, 0xc042137c80, 0xc042137ca0, 0xbf5fe0, 0xc04216e180, 0x0, 0x0)
	C:/Users/Paul/go/src/foobar/foobar-service/svc/transport_http.go:68 +0x2a8
foobar/foobar-service/svc/server.Run.func2(0xbf5fe0, 0xc04213e630, 0xf8c438, 0xc0420384a0, 0xc042137c40, 0xc042137c60, 0xc042137c80, 0xc042137ca0, 0x992e0a, 0x5, ...)
	C:/Users/Paul/go/src/foobar/foobar-service/svc/server/run.go:91 +0x1fa
created by foobar/foobar-service/svc/server.Run
	C:/Users/Paul/go/src/foobar/foobar-service/svc/server/run.go:94 +0x4f6
```

pauln:
As discussed in #209, moving to Gorilla Mux provides straightforward support for different HTTP verbs on the same path (thereby allowing for RESTful APIs to be generated).

Closes #209.


adamryman:
Keep going!

https://github.com/pauln/truss/pull/1

adamryman:
Question: in `$GOPATH/src/github.com/TuneLab/truss` does `make test` work? Not sure about windows.

This runs the integration tests, to make sure http transport works as we expect.

See here: https://github.com/TuneLab/truss/blob/master/DEVELOPING.md

pauln:
Unfortunately, `make test` doesn't work on Windows.  I might just switch to my Linux VM for dealing with the integration tests, as I don't really have time to try to port `make test` to Windows at the moment.

pauln:
The bulk of the failures were due to the generated HTTP requests using lower-case verbs.  Running them through ToUpper gets a lot further through the integration tests, but `TestGetWithPathParamsRequest_IncompletePath` fails as it's no longer relevant (as written).  The test is expecting a JSON error complaining about the number of parts in the path, but the malformed request now simply results in an HTTP 404 (Not Found).

A couple of options immediately spring to mind:

- Remove this test entirely, as it's now effectively testing how Gorilla Mux handles a (specific) bad request
- Update this test to check for a 404 response

If you let me know which approach (whether one of these or something else entirely) you'd prefer, I'm happy to implement it.

adamryman:
Unless you want to re-implement it, I say that it is fine to remove it entirely as I believe that gorilla mux would already test this.

This is awesome work, we have been hoping to get around to this. :100: 

I'll do a real review tomorrow morning. Looks great at a glance though. Should definitely be able to get this in before end of day tomorrow.

gengokit/httptransport/templates/server.go
```diff
@@ -49,79 +46,6 @@ var ServerDecodeTemplate = `
 // to fail. So don't change them purely because you think they look a little
 // funny.
 
-// PathParamsTemplate is a source code literal of the code for the PathParams
```
adamryman, line 52: It is great to see this code removed, maybe @hasLeland feels sad about it because he wrote it, though he is probably glad that we don't have to maintain it anymore.

pauln:
> I believe that gorilla mux would already test this

My thoughts exactly - consider it removed!

adamryman (approved):
I lied, I just reviewed it now.

So simple, such a small actual diff. Mostly removing code. (besides vendoring) :1st_place_medal: 

@pauln you are being talked about in the TuneLab office, we are excited to have a new contributer adding the features they care about.

   LGTM
 \ʕ◔ϖ◔ʔ/

(Besides either removing or rewriting the one failing integration test)

adamryman:
THEY PASS. :shipit:

gengokit/httptransport/templates/server.go
hasLeland: More code that we don't have to maintain is always an improvement!

Issue #211, opened 2017-07-27
Proposal: As a developer I want truss to be semantically versioned so I can version lock.


matthewhartstonge:
Hi TuneLab team!

First of all, thank you for truss!!! It's going to save us many hours of manual toil 👍👍👍
In the lead up to generating dependency management (as mentioned via MatthewHartstonge/truss#3) I was wondering if it would be possible to start versioning truss itself semantically to hold true to the paradigm being laid out with `go dep` which will be merged at some point into the standard go toolset.

I've started a PR that begins the journey into 'go dep~ism' but started with a focus on glide (what we have been using internally for all our go projects) but will change it over to `go dep` tomorrow, then begin the descent into truss'ing `go dep` dependency config file generation.

I would like to propose starting tagging at pre-release `v0.1.0` until it's potentially stable (as per the readme) for 2 reasons:
a) being semantically go dep~able
b) ease of takeup via github releases with pre-built binaries
c) SHA hashes make me 😭  😂

adamryman:
I do like this. Though I would want both a version and a sha. A sha for when truss has little changes that are not breaking, rather than bumping the version on every commit. Then the version for breaking changes.

PR #212, opened 2017-07-27
Feature/semantic versioning

matthewhartstonge:
Implements TuneLab/truss#211

As mentioned in #211, this currently needs to be re-worked to use `go dep` instead of glide.


Makefile
```diff
@@ -1,31 +1,34 @@
 # Makefile for Truss.
 #
-SHA := $(shell git rev-parse --short=10 HEAD)
 
-MAKEFILE_PATH := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
-VERSION_DATE := $(shell $(MAKEFILE_PATH)/commit_date.sh)
+# Pre-install required binaries
+$(shell go get github.com/golang/protobuf/protoc-gen-go)
```
adamryman, line 5: Do not do this. This means that whenever the make file is run these will be executed. Even when running tests.

Makefile
```diff
@@ -1,31 +1,34 @@
...
+VERSION_DATE := $(shell go-datefmt -ts ${GIT_COMMIT_EPOC} -fmt UnixDate -utc)
 
 # Build native Truss by default.
 default: truss
 
 dependencies:
-	go get github.com/golang/protobuf/protoc-gen-go
-	go get github.com/golang/protobuf/proto
-	go get github.com/jteeuwen/go-bindata/...
+	glide install
```
adamryman, line 19: These are binary dependencies for truss, not dependencies of a truss service.

adamryman (commented):
Hi @MatthewHartstonge.

If you want to change the default dependency management, please still commit all of the vendored code.

truss is a binary, and such we want to keep all of our vendored code in our repo, just in case those dependencies disappear.

If you want to make a good argument for removing them, we are up to hear it.

I would still want the git commit in the generated code as well. As we do not want to bump the semantic version for every commit we make, though we still want to know what the sha of the compiled version of truss was that generated code. As this allows us to know exactly what someone has installed.

In addition, please update the CI to work. You can get travis to run the ci on your fork if you want to test things.

Thanks for your interest! We look forward to seeing what you come up with.

matthewhartstonge:
Cheers for the feedback, will add to my todo list. 
For now I'm going to close this PR until I get time to deal with this issue again.

PR #215, opened 2017-07-31
Add logger to context

pauln:
Add go-kit logger to context in `MakeHTTPHandler` so that it's available for use in handlers/middlewares.

### DIFF
--- a/gengokit/httptransport/templates/server.go
+++ b/gengokit/httptransport/templates/server.go
@@ -103,6 +103,8 @@ func MakeHTTPHandler(ctx context.Context, endpoints Endpoints, logger log.Logger
 	{{- end }}
 	m := mux.NewRouter()
 
+	ctx = context.WithValue(ctx, "logger", logger)
+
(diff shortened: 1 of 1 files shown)


hasLeland:
We try to follow the practice of only putting things which are request scoped into the request context. That would include things like HTTP headers, but it excludes server-scoped items like loggers.

See this fantastic article by Peter Bourgon, a primary author of Go Kit, which delves into why and how we use context: https://peter.bourgon.org/blog/2016/07/11/context.html

The relevant quote:
> A database handle, or a logger, is generally created with the server, not with the request. So don’t use the context to pass those things around; instead, provide them to the middleware(s) that need them at construction time.

For this reason, `truss` will not add loggers to the context.

----
An alternative would be creating middlewares in a step which lets you pass in a logger. Or, the middlewares could instantiate their own loggers. What we do is instantiate a package-level logger in the `NewService` function of our truss services which can be accessed by other handlers or by middlewares.

PR #216, opened 2017-07-31
Feature/gokit v0.5.0 support

matthewhartstonge:
I noticed between starting this, there was a PR that moved middlewares into the handlers folder, which now causes a cyclic dependency.
Currently with this PR I've moved `middlewares.go` back out in order to remove the cycle, hence integration tests break. I believe this was to make it easier for the end user by having everything in `handlers/`?
What are your thoughts on removing this cycle?


hasLeland:
@MatthewHartstonge As far as I can tell from a brief look, the cyclic dependency is at its root due to the creation of the `{{GoName .Service.Name}}Server` interface within `handlers/handlers.go`.

That interface should probably go in a separate file in the same folder where the `.pb.go` file is located. That way it can be used in place of the `pb.NAMEServer` interface.

zaquestion:
GoKit potential upgrade is being covered here: https://github.com/TuneLab/truss/pull/224

PR #217, opened 2017-08-01
Update to stdlib log package

adamryman:
The idea is by updating to the stdlib log package, we get logging for the ports by default. And if someone wants to use the gokit logger, they can redirect the stdlib log to the go-kit logger.

For logs in handler/middlewares, the idea is that those are setup by the user of truss however they want.


adamryman:
Woops. I actually meant to submit this PR to my fork. This is a work in progress.

hasIan:
What does "logging for the ports" mean? I'm missing something here. Truss depends on go-kit, so I don't see why we wouldn't just use its logger.

adamryman:
Logging for the ports means that we still log the port values `"http" ":5040"`.

The reason is because the user of truss has no control how this logger is initialized if it is a go-kit logger. Though if we use the stdlib log package, then the user of truss can initalize a go-kit logger and redirect the standard lib logger to it.

I was running into this problem yesterday when trying to configure the logger in my own truss project. 

See: https://github.com/myambition/ambition/blob/9f5a0722a061f94a5e923117e5945539e461ea48/services/model/model-service/cmd/model-server/main.go#L23

zaquestion (approved):
This seems solid. Particularly when considering that main.go and by extension server.Run can be bypassed by creating your own version.

hasIan:
To clarify, Adam and I talked IRL after my question above and it makes sense--the logger that the generated Truss service uses is not exposed to the user anywhere, so it's incumbent upon the user to use his own logger. By switching to the stdlib logger, the user can redirect that to the logger of his choice, or ignore it. 👍

PR #224, opened 2017-08-03
Update truss to latest go-kit version through .pb.go rewrite

adamryman:
Being inspired by @pauln and @MatthewHartstonge efforts to move truss to a newer version of go-kit here is an attempt to get us to the latest go-kit without waiting for golang 1.9 type aliases.

The solution is pretty straightforward, rewrite the `.pb.go` files generated by `protoc-gen-go`
- [X] Add the import `newcontext "context` to every generated `.pb.go`
- [X] Update methods the `{{.Service.Name}}Server interface` in every `.pb.go` to take `newcontext.Context` as their first parameter

All other updates are modifying the truss's generated code to use `"context"` as well as updating little things that go-kit has changed since our freeze last year.

Here is the diff on `echo.pb.go` between current master and this PR.

```
 import (
        context "golang.org/x/net/context"
+       newcontext "context"
        grpc "google.golang.org/grpc"
 )
```
```
 type EchoServer interface {
        // Echo "echos" the incoming string
-       Echo(context.Context, *EchoRequest) (*EchoResponse, error)
+       Echo(newcontext.Context, *EchoRequest) (*EchoResponse, error)
        // Louder "echos" the incoming string with `Loudness` additional exclamation marks
-       Louder(context.Context, *LouderRequest) (*EchoResponse, error)
+       Louder(newcontext.Context, *LouderRequest) (*EchoResponse, error)
        // LouderGet is the same as Louder, but pulls fields other than Loudness (i.e. In) from query params instead of POST
-       LouderGet(context.Context, *LouderRequest) (*EchoResponse, error)
+       LouderGet(newcontext.Context, *LouderRequest) (*EchoResponse, error)
 }
```


zaquestion:
For the record, go 1.9 is expected to release this month.
https://tip.golang.org/doc/go1.9
> Go 1.9 is not yet released. These are work-in-progress release notes. Go 1.9 is expected to be released in August 2017.

adamryman:
Yep, very excited for that.

gengokit/handlers/templates/endpoint_middlewares.go
```diff
@@ -1,35 +0,0 @@
-package templates
-
-const EndpointsBase = `
```
hasLeland, line 3: Nice catch on the removal of the duped code :+1:

cmd/truss/main.go
```diff
@@ -269,6 +260,11 @@ func parseServiceDefinition(cfg *truss.Config) (deftree.Deftree, *svcdef.Svcdef,
 		return nil, nil, errors.Wrapf(err, "failed to create service definition; did you pass ALL the protobuf files to truss?")
 	}
 
+	// TODO: Remove once golang 1.9 comes out and type aliases solve context vs golang.org/x/net/context
+	if err := rewritePBGoForContext(sd.Service.Name, pbgoPaths); err != nil {
```
zaquestion, line 264: Is this call and the function itself all that need to be removed once 1.9 launches

adamryman: Correct. 

Though we should probably remove the function as well, just to remove unused code. Regardless, only the function call needs to be removed.

We will not actually _need_ to remove it, though it will be better form to not modify the generated `.pb.go` files if we do not need to.

zaquestion (approved):

matthewhartstonge:
Thankyou for your work on this team :) we are very grateful for your efforts and saving us from a lot of boilerplate!

PR #226, opened 2017-08-08
Only process {svcname}Server interfaces in svcdef.New

pauln:
Also logs unexpected interfaces (i.e. oneof interfaces) so that the user can see that they were skipped.  This prevents the panic mentioned in #225, but does not add actual support for `oneof` fields.

PR #227, opened 2017-08-09
Proposal: Use jsonpb to (un)marshall messages to/from JSON

pauln:
The stdlib encoding/json package isn't able to (un)marshal some advanced protobuf features such as `oneof` fields.  [golang/protobuf/jsonpb](http://github.com/golang/protobuf/jsonpb) should be used instead, as it follows the [canonical proto3 JSON encoding](https://developers.google.com/protocol-buffers/docs/proto3#json) and therefore supports these features.

This change should allow `oneof` fields to be used, though the warning message added by #226 can probably be removed again (or degraded to debug level) as it (mostly) relates to the previous lack of support for `oneof` fields.

### DIFF
--- a/cmd/_integration-tests/transport/http_test.go
+++ b/cmd/_integration-tests/transport/http_test.go
@@ -17,6 +17,7 @@ import (
 	pb "github.com/TuneLab/truss/cmd/_integration-tests/transport/transportpermutations-service"
 	httpclient "github.com/TuneLab/truss/cmd/_integration-tests/transport/transportpermutations-service/svc/client/http"
 
+	"github.com/golang/protobuf/jsonpb"
 	"github.com/pkg/errors"
 )
 
@@ -606,7 +607,21 @@ func testHTTP(
 		t.Fatal(errors.Wrap(err, "cannot make http request"))
 	}
 
-	err = json.Unmarshal(respBytes, &resp)
+	switch v := resp.(type) {
+	case *pb.GetWithQueryResponse:
+		err = jsonpb.UnmarshalString(string(respBytes), v)
+	case *pb.GetWithRepeatedQueryResponse:
+		err = jsonpb.UnmarshalString(string(respBytes), v)
+	case *pb.GetWithEnumQueryResponse:
+		err = jsonpb.UnmarshalString(string(respBytes), v)
+	case *pb.PostWithNestedMessageBodyResponse:
+		err = jsonpb.UnmarshalString(string(respBytes), v)
+	case *pb.MetaResponse:
+		err = jsonpb.UnmarshalString(string(respBytes), v)
+	default:
+		t.Fatalf("Unknown response type: %T", v)
+	}
+
 	if err != nil {
 		t.Fatal(errors.Wrapf(err, "json error, got response: %q", string(respBytes)))
 	}
--- a/gengokit/httptransport/templates/server.go
+++ b/gengokit/httptransport/templates/server.go
@@ -14,7 +14,11 @@ var ServerDecodeTemplate = `
 			return nil, errors.Wrapf(err, "cannot read body of http request")
 		}
 		if len(buf) > 0 {
-			if err = jsonpb.UnmarshalString(string(buf), &req); err != nil {
+			// AllowUnknownFields stops the unmarshaler from failing if the JSON contains unknown fields.
+			unmarshaller := jsonpb.Unmarshaler{
+				AllowUnknownFields: true,
+			}
+			if err = unmarshaller.Unmarshal(bytes.NewBuffer(buf), &req); err != nil {
 				const size = 8196
 				if len(buf) > size {
 					buf = buf[:size]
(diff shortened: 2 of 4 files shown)


cmd/_integration-tests/transport/http_test.go
```diff
@@ -606,7 +607,21 @@ func testHTTP(
 		t.Fatal(errors.Wrap(err, "cannot make http request"))
 	}
 
-	err = json.Unmarshal(respBytes, &resp)
+	switch v := resp.(type) {
+	case *pb.GetWithQueryResponse:
```
zaquestion, line 611: Can these be combined into 1 case statement? 
```
 case *pb.GetWithQueryResponse, *pb.GetWithRepeatedQueryResponse,...:
```

adamryman: You cannot:
```
./http_test.go:612: cannot use v (type interface {}) as type proto.Message in argument to jsonpb.UnmarshalString:
        interface {} does not implement proto.Message (missing ProtoMessage method)
```

By being only one type per case then the type of `v` is known. If you combine them into one then `v` is of type `interface{}`.

pauln: Yeah, I found that out the hard way - tried combining cases, still got an interface.  It's more verbose (and repetitive) than I'd like, but as `jsonpb.Unmarshal` wants a `proto.Message`, this was the shortest / simplest way I could find to do it.

pauln:
I've just pushed another commit, as I discovered a regression wherein unknown keys in the request JSON resulted in an error from `jsonpb.UnmarshalString`, though they were previously silently ignored.  The workaround requires switching to a `jsonpb.Unmarshaler` so that we can set `AllowUnknownFields`.

gengokit/httptransport/templates/server.go
```diff
@@ -173,9 +180,12 @@ func (h httpError) Headers() http.Header {
 // EncodeHTTPGenericResponse is a transport/http.EncodeResponseFunc that encodes
 // the response as JSON to the response writer. Primarily useful in a server.
 func EncodeHTTPGenericResponse(_ context.Context, w http.ResponseWriter, response interface{}) error {
-	encoder := json.NewEncoder(w)
-	encoder.SetEscapeHTML(false)
```
adamryman, line 177: We are a little concerned about this. Could you point us at documentation that jsonpb, does *not* do html escaping? 

See here https://godoc.org/encoding/json#Encoder.SetEscapeHTML. We had a problem with this in production before.

We will be testing this in our staging environment for one of products here at Tune, as we do not have tests for the HTML escaping in json. I believe that I'll be able to do that before the end of the week. If you would like to add a test for HTML characters being escaped in json, it would be highly appreciated and give strong confidence for merging this.

pauln: A quick test shows that it unfortunately _is_ doing html escaping.  I'll do a bit of digging to see if that's able to be changed.

pauln: I can't see a way to disable the HTML escaping (`jsonpb` internally uses `encoding/json`, but doesn't currently have a way to call `SetEscapeHTML`), so I've filed an issue.

adamryman (dismissed):
I'm putting a block on this PR till we resolve `encoder.SetEscapeHTML(false)` being removed and not having an equivalent in `jsonpb`.

[See this review in this PR](https://github.com/TuneLab/truss/pull/227/files#r132371930)
[See this issue in golang/protobuf](https://github.com/TuneLab/truss/pull/227/files#r132371930)

pauln:
I've [submitted a PR](https://github.com/golang/protobuf/pull/409) which adds a `SkipEscapeHTML` option to `jsonpb.Marshaler` which should unblock this if/when merged.  It currently fails CI on Go1.6 due to lack of options on `json.Encoder`, so I'm not sure how soon they might be willing to merge it.

zaquestion:
With truss now able to support `google.protobuf` types through gogo, these changes seem more relevant than ever!

For
```
message SearchResponse {
  int64 num_results = 1;
  repeated google.protobuf.Struct results = 2;

}
```

On master trying to marshal the `google.protobuf.Struct` type yields
```
$ curl 'localhost:5050/search' -d '{ "query": "foo" }'
{"num_results":1,"results":[{"fields":{"displayName":{"Kind":{"StringValue":"Denver, CO"}},"objectType":{"Kind":{"StringValue":"place"}}}}]}
```
regenerated with these changes merged in we get the clean response we want.
```
$ truss service.proto
$ curl 'localhost:5050/search' -d '{ "query": "foo" }'
{"numResults":"1","results":[{"displayName":"Denver, CO","objectType":"place"}]}
```
The response data is hardcoded and not actually based on the query or anything. I haven't tried in the other direction, but based on the code here and the above value add already, I think its worth pushing through. The original hold on escape html was more specific to TUNE and those services aren't in production any more, I'm on board with starting fresh.

lelandbatey:
The original HTML escape case that we where hung up on is something we don't really care about any longer. The improvements from `jsonpb` vastly outweigh the edge cases, so I'm very in favor of merging this PR.

zaquestion (approved):
Approved in conjunction with @lelandbatey

zaquestion:
@pauln After all this time, we've come to our senses! Excited to have this in :D

PR #229, opened 2017-08-23
Update the bundled protobuf descriptor.proto to 3.4.0

jackwakefield:
The update adds new options such as `php_namespace`.


adamryman:
LGTM, thanks for becoming a contributor! 

I'm curious, what encouraged you to update this bundled file? Something with `php`?

jackwakefield:
Exactly that, `php_namespace` specifically.

Protobuf 3.3 released back in May was the first GA release for PHP, the bundled `descriptor.proto` predated that so was missing a few options.

PR #230, opened 2017-09-02
Add support for custom HTTP verbs

### LINKED ISSUE #218: truss does not support http verb HEAD or custom http verbs
Currently truss does not support adding a http HEAD endpoint or any other custom http verb.

The google protobuf http annotations has support for this feature. Though truss currently is not setup to parse it.

https://github.com/TuneLab/truss/blob/master/deftree/googlethirdparty/http.proto#L263

hasLeland:
Implements / Fixes #218 

## What does this add?

This PR adds support for using custom HTTP verbs in your routes for RPC methods. For example, you can now define a method as responding to `HEAD` requests, which in code would be like so:

```
service ExmplService {
  rpc ExmplMethod (RequestStrct) returns (ResponseStrct) {
    option (google.api.http) = {
      custom {                              ┐
        kind: "HEAD"                        │ This newly supported syntax is how
        path: "/foo/bar/{SomeFieldName}"    │ you define a custom HTTP verb.
      }                                     ┘
      // This 'body' field is optional
      body: "*"
    };
  }
}
```

You can define additional custom HTTP verbs using the `additional_bindings` syntax. For example:

```
service ExmplService {
  rpc ExmplMethod (RequestStrct) returns (ResponseStrct) {
    option (google.api.http) = {
      post: "/foo/bar/{SomeFieldName}"
      body: "*"
      
      additional_bindings {
        custom {
          kind: "FIRSTVERB"
          path: "/foo/bar/{SomeFieldName}"
        }
      }
      additional_bindings {
        custom {
          kind: "SECONDVERB"
          path: "/foo/bar/{SomeFieldName}"
        }
      }
    };
  }
}
```

## What does this *not* add?

This addition does not add the ability to differentiate between different HTTP verbs within a truss handler function. What does this mean? Let's take the service defined in the second example above, the one with multiple HTTP bindings for a single RPC method. The way truss works, you would get a single handler function in the generated `handlers/handlers.go` file, a handler function called `ExmplMethod`. Now, if a client makes a `POST` request to the `ExmplMethod` route, it'll be handled in the `ExmplMethod` function in `handlers/handlers.go`; the same things happen if you make a `FIRSTVERB` request to the `ExmplMethod` route, that is, the logic for how to respond to that route will be handled by the `ExmplMethod` function in `handlers/handlers.go`.

However, the function `ExmplMethod` in `handlers/handlers.go` *currently* does not have any way of knowing whether a request from a client was made using the `POST` verb or whether is was made using the `FIRSTVERB` verb. This is the remaining deficiency of this implementation, that currently the handler has no way of knowing which verb was used for a particular RPC. We welcome discussion on ways to expose this information to the handler in a clean and idiomatic way.

## Changelog

- [X] Expand the protobuf parser in `deftree/svcparse/` to correctly parse the `CustomHTTPPattern` syntax defined in `googlethirdparty/http.proto`
- [X] Modify the `svcparse/` package to leverage the new `CustomHTTPPattern` information from the protobuf parser
- [x] Update documentation generation
- [x] Add integration tests for verification of custom HTTP verb support

*Note: @adamryman has edited this PR text as it has become a WIP*

### DIFF
--- a/deftree/svcparse/parser.go
+++ b/deftree/svcparse/parser.go
@@ -83,6 +83,8 @@ type Method struct {
 // HTTPBinding holds information extracted by the parser about each HTTP
 // binding within each method.
 type HTTPBinding struct {
+	// At this time, the way to provide a "description" of an HTTP binding is
+	// to write a comment directly above the "option" statement in an rpc.
 	Description string
 	Fields      []*Field
 	// CustomHTTPPattern contains the fields for a `custom` HTTP verb. It's
@@ -141,7 +143,7 @@ type HTTPBinding struct {
 // HTTP binding.
 type Field struct {
 	// Name acts as an 'alias' for the Field. Usually, it has the same value as
-	// "Kind", though there are no guaruntees that they'll be the same. Name
+	// "Kind", though there are no guarantees that they'll be the same. Name
 	// should never be used as part of "business logic" it is purely as a
 	// human-readable decorative field.
 	Name        string
--- a/deftree/svcparse/parser.go
+++ b/deftree/svcparse/parser.go
@@ -85,11 +85,65 @@ type Method struct {
 type HTTPBinding struct {
 	Description string
 	Fields      []*Field
+	// CustomHTTPPattern contains the fields for a `custom` HTTP verb. It's
+	// name comes from the name for this construct in the http annotations
+	// protobuf file. The following is an example of a protobuf service with
+	// custom http verbs and the equivelent HTTPBinding literal:
+	//
+	// Protobuf code:
+	//
+	// service ExmplService {
+	//   rpc ExmplMethod (RequestStrct) returns (ResponseStrct) {
+	//     option (google.api.http) = {
+	//       custom {
+	//         // The verb itself goes in the "kind" field
+	//         kind: "MYVERBHERE"
+	//         // Likewise, path goes in the "path" field. As always, the path
+	//         // may have parameters within it.
+	//         path: "/foo/bar/{SomeFieldName}"
+	//       }
+	//       // This 'body' field is optional
+	//       body: "*"
+	//     };
+	//   }
+	// }
+	//
+	// Resulting HTTPBinding:
+	//
+	// HTTPBinding{
+	//     Fields: []*Field{
+	//         &Field{
+	//             Description: "// This 'body' field is optional\n",
+	//             Name:  "body",
+	//             Kind:  "body",
+	//             Value: "*",
+	//         },
+	//     },
+	//     CustomHTTPPattern: []*Field{
+	//         &Field{
+	//             Description: "// The verb itself goes in the \"kind\" field\n",
+	//             Name:        "kind",
+	//             Kind:        "kind",
+	//             Value:       "MYVERBHERE",
+	//         },
+	//         &Field{
+	//             Description: "// Likewise, path goes in the \"path\" field. As always, the path\n\t\t\t\t\t// may have parameters within it.\n",
+	//             Name:  "path",
+	//             Kind:  "path",
+	//             Value: "/foo/bar/{SomeFieldName}",
+	//         },
+	//     },
+	// },
+	CustomHTTPPattern []*Field
 }
 
 // Field holds information extracted by the parser about each field within each
 // HTTP binding.
 type Field struct {
+	// Name acts as an 'alias' for the Field. Usually, it has the same value as
+	// "Kind", though there are no guaruntees that they'll be the same. Name
+	// should never be used as part of "business logic" it is purely as a
+	// human-readable decorative field.
 	Name        string
 	Description string
 	Kind        string
@@ -283,11 +337,12 @@ func ParseHttpBindings(lex *SvcLexer) ([]*HTTPBinding, error) {
 		if err != nil {
 			return nil, err
 		}
-		fields, err := ParseBindingFields(lex)
+		fields, custom, err := ParseBindingFields(lex)
 		if err != nil {
 			return nil, err
 		}
 		new_opt.Fields = fields
+		new_opt.CustomHTTPPattern = custom
 		good_position := lex.GetPosition()
 
 		tk, val = lex.GetTokenIgnoreWhitespace()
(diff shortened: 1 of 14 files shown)


deftree/svcparse/parser.go
```diff
@@ -85,11 +85,65 @@ type Method struct {
...
+	//     },
+	// },
+	CustomHTTPPattern []*Field
 }
 
 // Field holds information extracted by the parser about each field within each
 // HTTP binding.
 type Field struct {
+	// Name acts as an 'alias' for the Field. Usually, it has the same value as
+	// "Kind", though there are no guaruntees that they'll be the same. Name
```
zaquestion: guarantee*

deftree/svcparse/parser.go
```diff
@@ -305,7 +360,7 @@ func ParseHttpBindings(lex *SvcLexer) ([]*HTTPBinding, error) {
 				rv = append(rv, more_bindings...)
 				good_position = lex.GetPosition()
 			} else if tk == EOF || tk == ILLEGAL {
-				return nil, parseErr("non-illegal token while parsing HttpBindings", lex.GetLineNumber(), fmt.Sprintf("(%v) of type %v", val, tk))
+				return nil, parseErr("legal token while parsing HttpBindings", lex.GetLineNumber(), fmt.Sprintf("(%v) of type %v", val, tk))
```
zaquestion: It seems like this error message should be indicating that EOF or an Illegal token is expected? Perhaps the parser is looking to finish its work. "legal token" doesn't tell me much and I might just as quickly assume that this was a typo and that illegal was meant here.

zaquestion (commented):
I see the test for including comments around the custom verbs. Do these comments make it into the docs?

deftree/svcparse/parser.go
hasLeland: That line will actually generate an error with the value:

    parser expected legal token while parsing HttpBindings in line '##', instead found '(BAD_THING) of type EXAMPLE'

which I feel is a relatively descriptive error message.

hasLeland: As to your review question about whether the comments make their way into the docs: No, no they do not. You see, I forgot to check how this affects documentation generation and it is only because of your comment that I'm checking now. Unfortunately, the docs generation for this is pretty broken, something I had forgotten about.

So I'm going to extend documentation generation to support this new addition, push the changes, and update this PR when I do.

adamryman: @hasLeland I think @zaquestion has a point about the code being confusing even if it produces a good error message. `parseErr` as a function also implies to me that is going to parse an error rather than return a error.

Maybe a rename to `parserErr` the extra `r` makes a difference to me in understanding what it does.

-------------------------------------------------------------
Extra:

I wrote this out for making a type for parse error, so I am going to post it, though it may be more than we need.
```
type parseErr struct {
	expected string 
	line int 
	val string
}

func (p parseErr) Error() string {
    return fmt.Errorf("parser expected %v in line '%v', instead found '%v'", p.expected, p.line, p.val)
}

...
return nil, parseErr {
   expected: "legal token while parsing HttpBindings", 
   line: lex.GetLineNumber(), 
   val: fmt.Sprintf("(%v) of type %v", val, tk)),
}
```

deftree/svcparse/parser.go
```diff
@@ -336,9 +476,9 @@ func ParseHttpBindings(lex *SvcLexer) ([]*HTTPBinding, error) {
 	return nil, optErr
 }
 
-func ParseBindingFields(lex *SvcLexer) ([]*Field, error) {
-
+func ParseBindingFields(lex *SvcLexer) ([]*Field, []*Field, error) {
 	rv := make([]*Field, 0)
+	var custom []*Field
```
adamryman: Why is the constructed differently than `rv`?

```
rv := make([]*Field, 0)
```

deftree/svcparse/parser.go
```diff
@@ -336,9 +476,9 @@ func ParseHttpBindings(lex *SvcLexer) ([]*HTTPBinding, error) {
 	return nil, optErr
 }
 
-func ParseBindingFields(lex *SvcLexer) ([]*Field, error) {
-
+func ParseBindingFields(lex *SvcLexer) ([]*Field, []*Field, error) {
```
adamryman: Possibly update to named return values to disambiguate the `[]*Field`s?

```
func ParseBindingFields(lex *SvcLexer) (fields []*Field, custom []*Field, err error) {
```

deftree/svcparse/parser.go
```diff
@@ -360,29 +504,54 @@ func ParseBindingFields(lex *SvcLexer) ([]*Field, error) {
 			lex.UnGetToken()
 			break
 		}
+		// Use recursion to parse custom HTTP verb sections
+		if val == "custom" {
+			err := fastForwardTill(lex, "{")
+			if err != nil {
+				return nil, nil, err
+			}
+			c, _, err := ParseBindingFields(lex)
```
adamryman: Can you not have custom within custom? Not sure of the significance of ignoring the middle param.

deftree/svcparse/parser.go
```diff
@@ -360,29 +504,54 @@ func ParseBindingFields(lex *SvcLexer) ([]*Field, error) {
...
+			if err != nil {
+				return nil, nil, err
```
adamryman: Would wrapping these errors be helpful? 
```
errors.Wrap(err, "error parsing custom binding field")
```

deftree/svcparse/parser_test.go
```diff
@@ -573,3 +581,149 @@ service FlowCombination {
...
+	r := strings.NewReader(customVerbWithComments)
+	lex := NewSvcLexer(r)
+	_, err := ParseService(lex)
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+// Test that parsing of a custom HTTP pattern returns a correct and expected HTTPBinding struct.
+func TestCustomHTTPPatternOutputExample(t *testing.T) {
```
adamryman, line 673: Lots of tests :+1:

adamryman (commented):
I believe integration tests are in order. I would say both a http transport integration test and a cli integration test.

deftree/svcparse/parser.go
hasLeland: No, you cannot have a custom within a custom, which is why I underscore away what would be the "custom" fields (the middle param).

deftree/svcparse/parser.go
hasLeland: No reason at all, will update to make these both more consistent.

deftree/svcparse/parser.go
hasLeland: Doing

deftree/svcparse/parser.go
hasLeland: I don't know if it would be materially helpful, but it certainly makes things more consistent.

adamryman (approved):

adamryman:
See #243

PR #239, opened 2017-09-22
Remove --pbout, ensure .pb.go are generated next to .proto

### LINKED ISSUE #236: Proposal: Always generate .pb.go next to original .proto.
- [x] Update default location of `.pb.go` to be in the same directory as the .proto definition.
- [x] Remove flag `--pbout` 

This will update truss such that `.pb.go` files generated are in line with what the rest of the `protoc-gen-go` users are doing. 

See related comments in [#201.](https://github.com/TuneLab/truss/issues/201#issuecomment-317564509)
> - Default truss behavior of putting generated `.pb.go` files into a different directory than the `.proto` files that defined them. This confuses `protoc-gen-go` which truss uses to generate the `.pb.go` files.
> - Proto packages imported in a `.proto` file used with `truss` must be imported by their full path relative to `$GOPATH/src`. This is so that the generated go code will have a proper import path that is buildable.

hasLeland:
Implements / Fixes #236

## What does this change?

This change removes the `--pbout` CLI flag for the `truss` binary, and changes where `truss` places the `.pb.go` files it generates. Instead of placing the `.pb.go` file(s) into the newly-created `NAME-service` directory with your other generated code, now `truss` places the `.pb.go` file(s) into the same directory as the `.proto` files which they're generated from.

To better illustrate what those changes mean, let's walk through the old behavior and the new behavior. Let's say you have a directory like this:

```
$ tree .
.
└── example.proto
```

First, let's look at the old behavior:

```
$ truss example.proto
$ tree -L 2
.
├── example.proto
└── example-service
    ├── cmd
    ├── docs
    ├── example.pb.go  <-  Notice that the .pb.go is placed
    ├── handlers           inside of the generated service
    └── svc
```

As you can see, the `example.pb.go` file is placed within the `example-service` directory generated by `truss`. Now, let's see how this differs from the new behavior, which places the `.pb.go` files in the same directory as the `.proto` files which the `.pb.go` files are generated from.

```
$ truss example.proto
$ tree -L 2
.                          Notice that the .pb.go file is now placed outside of
├── example.pb.go      <-  generated service directory, alongside the existing
├── example.proto          .proto file which defines the service.
└── example-service
    ├── cmd
    ├── docs
    ├── handlers
    └── svc
```
In the above, we can see the new behavior which places the `.pb.go` file next to the `.proto` file defining the service.

- [X] Remove the `--pbout` flag from `cmd/main.go`
- [X] Remove the integration tests for `--pbout`
- [X] Change `.pb.go` output to be the same as the `.proto` output


cmd/_integration-tests/cli/cli_test.go
```diff
@@ -191,7 +179,7 @@ func TestAdditionalBindings(t *testing.T) {
 
 func testEndToEnd(defDir string, subcmd string, t *testing.T, trussOptions ...string) {
 	path := filepath.Join(basePath, defDir)
-	err := createTrussService(path)
+	err := createTrussService(path, trussOptions...)
```
adamryman, line 182: Wait... Was this never testing the `--pbout flag`? That is kinda bad... Oh well.

cmd/truss/main.go
```diff
@@ -199,20 +198,27 @@ func parseInput() (*truss.Config, error) {
...
-		cfg.PBPackage = p.ImportPath
-		cfg.PBPath = p.Dir
+	//
+	// It used to be the case that the package where the .pb.go files would be
+	// placed by default was into the newly-generated truss service. However,
+	// we now want to keep our .pb.go files directly next to our .proto files,
+	// so that's where we're going to place them by default. This implies an
+	// assumption that our .proto files exist within our GOPATH.
+	//
+	// Also, if they've passed in multiple protobuf files, we're going to use
```
adamryman, line 208: In the future, we want to support passing in several protobuf `packages`. Though only one package will be allowed to have a `Service` definition. 

Is there a way to update this such that `protoDir` is created from the `DefPath` of the `.proto` file that contains the service definition? I know you have that function that tells you what the `ServiceName` is, so I believe that function could be used/modified to figure out what the path to the `.proto` file that contains the service definition.

That make sense? It was a bit wordy.

adamryman (commented):
Looks good to me. Though please investigate my comment about using the service definition path. It would be nice to roll that in here.

cmd/_integration-tests/cli/cli_test.go
hasLeland: It is kinda bad, though I did fix these tests on master to see if things where broken and they where not.

cmd/truss/main.go
adamryman: I think this can be added in a future PR.

adamryman (approved):

Issue #242, opened 2017-11-03
Proposal: Update truss to generate .pb.go files even when gRPC service is not defined.


adamryman:
If I have a `.proto` file that contains `message`s that I would like to use in multiple services:
```
$ cat shared.proto
syntax = "proto3";

package shared;

message SharedRequest {
  string In = 1;
}
```

And if `truss` is run on this file, **I propose that truss generates a corresponding `.pb.go` file.** 

In the case of `shared.proto` the generated `shared.pb.go` would look like [this](https://gist.github.com/adamryman/2ab35e1dadb0cfbf9e0a9115a2cef404).

Currently, truss panics because it cannot find the service name. (This used to have a nice error, guess this needs to have a test as well)

```
$ truss shared.proto
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x64568f]
```

To implement:
- [x] Review and merge #239 
- [x] fix panic caused here [parsesvcdef.go:66](https://github.com/TuneLab/truss/blob/a9431ca6707d745dfb058a2af09ec052eea11281/truss/parsesvcname/parsesvcname.go#L66), return error no `Service` or `Service.Name` 
- [x]  [`cmd/truss/main.go`, function `parseInput()`](https://github.com/TuneLab/truss/blob/a9431ca6707d745dfb058a2af09ec052eea11281/cmd/truss/main.go#L146) needs to 
    - [x] handle the error from `parsesvcname.FromPaths`
    - [x] warn that no service is defined
    - [x] info that `.pb.go` files will be generated
    - [x] be wired up to `execprotoc.GeneratePBDotGo()` to actually generate the .pb.go files.

This is a step to fix #241

adamryman:
Fixed by #246.

Thanks again @eriktate and welcome to being a contributor!

PR #246, opened 2017-12-13
Support generating .pb.go files from protobufs without service definitions

eriktate:
Effectively just making sure protoc is still run against .proto files that define shared messages and no RPC service.

This PR addresses #242


truss/config.go
```diff
@@ -18,4 +18,7 @@ type Config struct {
 	DefPaths []string
 	// The files of a previously generated service, may be nil
 	PrevGen map[string]io.Reader
+
+	// The flag that specifies if a service should be built.
+	NoService bool
```
eriktate: If it would be preferable for `parseInput` to return an additional value (`NoService`) rather than extending the `Config` struct I can rework that.

adamryman:
Hey @eriktate, I'm going to review this in the next few hours. I'm sorry for causing a conflict.

I'll PR a fix on top of your PR, to resolve the conflict. We just wanted the 1.9 fix in very strongly. 

Though your PR came up first, so it is only polite for me to resolve this conflict I created.

Thanks for the PR! I look forward to taking a closer look very shortly.

adamryman:
Hey @eriktate, here is that PR I mentioned: https://github.com/eriktate/truss/pull/3

Changes look good! Sorry that some of the same logic was reworked right after you put this PR up.

Hope to talk with you soon.

eriktate:
@adamryman No worries! Just merged your PR resolving the conflicts, so this should be good to go!

adamryman:
Just because I make some extra changes in the PR on your PR, I would like one other team member to take a look. 

@zaquestion, @lelandbatey, if either of you could give a :white_check_mark: then I think we will be ready to merge.

adamryman (approved):

lelandbatey (approved):
Yay for the fix toward #242!

PR #247, opened 2017-12-15
Update svcout handling for 1.9+

adamryman:
In golang 1.9 `go/build` package changed how `build.Import` works.

This change removes dependencies on undocumented stdlib features and
keeps `svcout` working as it did previously.

### DIFF
--- a/cmd/truss/main.go
+++ b/cmd/truss/main.go
@@ -225,7 +225,7 @@ func parseInput() (*truss.Config, error) {
 	return &cfg, nil
 }
 
-// parseSVCOout handles the difference between relative paths and go package
+// parseSVCOut handles the difference between relative paths and go package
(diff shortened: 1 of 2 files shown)


cmd/truss/main.go
```diff
@@ -225,6 +225,15 @@ func parseInput() (*truss.Config, error) {
 	return &cfg, nil
 }
 
+// parseSVCOout handles the difference between relative paths and go package
```
zaquestion: parseSVCOut*

zaquestion (commented):
Everything else looks good

zaquestion (approved):

lelandbatey:
I'll also chime in to say that this PR seems solid, glad this is being fixed for 1.9!

PR #249, opened 2018-01-16
fix for repeated enums; test for repeated enums

JohnF42:
Repeated enums are currently creating invalid code. This fixes this and adds a test.

### DIFF
--- a/cmd/_integration-tests/cli/test-service-definitions/2-repeated/repeated.proto
+++ b/cmd/_integration-tests/cli/test-service-definitions/2-repeated/repeated.proto
@@ -20,6 +20,13 @@ service TEST {
   }
 }
 
+enum TestType {
+  TEST1 = 0;
+  TEST2 = 1;
+  TEST3 = 2;
+}
+
+
 message RepeatedTypeRequest {
   repeated double A = 1;
   repeated float B = 2;
@@ -35,6 +42,7 @@ message RepeatedTypeRequest {
   repeated bool L = 12;
   repeated string M = 13;
   repeated bytes N = 14;
+  repeated TestType O = 15;
 }
 
 
@@ -53,5 +61,6 @@ message RepeatedTypeResponse {
   repeated bool L = 12;
   repeated string M = 13;
   repeated bytes N = 14;
+  repeated TestType O = 15;
 }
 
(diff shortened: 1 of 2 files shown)


zaquestion:
@lelandbatey @adamryman Can you take a look? Are there implications around parsing the request here?

cmd/_integration-tests/cli/test-service-definitions/2-repeated/repeated.proto
```diff
@@ -20,6 +20,13 @@ service TEST {
   }
 }
 
+enum TestType {
```
adamryman, line 23: I would prefer `EnumType` or `EnumTestType`.

gengokit/httptransport/httptransport.go
```diff
@@ -276,7 +276,7 @@ func createDecodeConvertFunc(f Field) (string, bool) {
 		needsErrorCheck = false
 	}
 
-	if f.IsEnum {
+	if f.IsEnum && !f.Repeated {
```
adamryman, line 279: Consider adding a comment when you run across something like this that might be unexpected.

cmd/_integration-tests/cli/test-service-definitions/2-repeated/repeated.proto
```diff
@@ -20,6 +20,13 @@ service TEST {
...
+  TEST1 = 0;
+  TEST2 = 1;
+  TEST3 = 2;
+}
+
+
```
adamryman, line 29: Possibly unneeded whitespace.

adamryman (approved):
Great! :100: :1st_place_medal: :fireworks: 

@JohnF42 the comments I left are for future considerations. I believe they would improve things, though the functionality you added is just fine without them.

adamryman:
@zaquestion this will have no implications around parsing the request. As we only support Enums in body of a request, and the json parser will handle that for us.

@JohnF42 great catch here! We were not testing repeated Enums and so we did not generate the "slice/array" code in the case the Enum was repeated.

Thanks for this addition, and especially than you for the test.

zaquestion:
@adamryman You sure about that? https://github.com/TuneLab/truss/pull/176/files

adamryman:
@zaquestion hmmm.

Could put the warning back in for now if someone tries to put a repeated Enum in the path or query.

Chaninging this:
https://github.com/TuneLab/truss/blob/master/gengokit/httptransport/httptransport.go#L115
```  
        if newField.IsEnum {
            continue
        }
```
to this
```
        if newField.IsEnum && !newField.Repeated {
            continue
        }
```

PR #255, opened 2018-05-19
example: update migrated repo reference

rdcx:


zaquestion:
Thanks for fixing that, there something wrong with the build which I'm hoping to look closer at this weekend. If you're running into more errors is probably that.

PR #258, opened 2018-05-27
Use goreleaser to manage releases

zaquestion:
Will build truss for Mac (+push to homebrew tap) and linux. Will add binaries and a generated changelog to the github release

Need to generate a GITHUB_TOKEN for TUNE to do the releases.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,50 @@
+project_name: truss
+release:
+  github:
+    owner: tuneinc
+    name: truss
+  name_template: '{{.Tag}}'
+brew:
+  description: "Truss helps you build go-kit microservices without having to worry about writing or maintaining boilerplate code"
+  homepage: "https://github.com/tuneinc/truss"
+  github:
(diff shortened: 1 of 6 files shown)


zaquestion:
@adamryman @lelandbatey There isn't a great way to install `protoc-gen-go` and other go project dependencies with brew and more so with goreleaser's formula generator. I considered opening a feature request on goreleaser to include support for the `resource` definition for brew formulas, but I don't think it would help much because protoc-gen-go doens't have released binaries anyways.

Beyond brew the release binaries this would generate also wouldn't have these dependencies. Thoughts on having `truss` check for these dependencies at runtime and offer to `go get` them for the user. This could be useful for `go get`ing the truss repo itself which contains the `annontations.proto` users reference.

zaquestion:
Had to create a new branch in this repo since this was originally coming from the tuneinc/truss repo which no longer exists, ironically because it became this repo. New PR here https://github.com/metaverse/truss/pull/275

PR #259, opened 2018-05-28
remove protoc-gen-truss-protocast from windows install script.

matthewhartstonge:
Fixes windows installs


zaquestion (approved):
Ah, I missed that.

PR #263, opened 2019-01-17
Update truss deps to be managed with modules

zaquestion:


zaquestion:
@adamryman works with go1.10 because `go mod vendor` keeps the vendors in sync.

kainoaseto:
Added commit for using drop in replacement of go-bindata that is more maintained and updated makefile to pull modules when modules are specified to be used.

ianfoo (commented):
LGTM, though I'm not up to date on Travis: what's up with 1.11.x being in both in the `go` section and the `matrix` session. Is `matrix` just additive to what's in `go`, setting up an env var for that specific version?

Apologies for not looking this up on Travis--but I know if I descend down that rabbit hole, I'll be reading about Travis and CI for at least a half hour.

zaquestion:
@ianfoo correct its additive as seen here: https://travis-ci.org/metaverse/truss/builds/483689196

This is intentional so that go1.11 is tested with and without GO111MODULE=on

PR #264, opened 2019-01-24
Update go-bindata ref and add support for modules

kainoaseto:
Added support for modules and go-bindata drop in replacement until a better solution can be added in.


Makefile
```diff
@@ -9,10 +9,13 @@ VERSION_DATE := $(shell $(MAKEFILE_PATH)/commit_date.sh)
 default: truss
 
 dependencies:
+ifeq ($(GO111MODULE), on)
+	go mod download
+endif
```
zaquestion: Shouldn't need this block since the below block will download the deps for you and install them.

.travis.yml
```diff
@@ -8,6 +8,11 @@ go:
   - 1.11.x
```
zaquestion, line 8: You can remove this here since you've got it in the matrix now. Otherwise you'll get a dup job (possibly intentional?)
Seen here: https://travis-ci.org/metaverse/truss/builds/483689196

Can probably also remove tip and put it in the below matrix as well

zaquestion (changes requested):
Looking for a response to the multiple `go1.11` jobs comment and probably dropping `go mod download` from the `Makefile` unless its actually needed.

Makefile
kainoaseto: Ah ok cool, I'd prefer to avoid go get entirely if all the other deps in the go.mod file are direct deps of the one's below. I'll work on getting these in there so `go mod download` is all that is needed.

.travis.yml
kainoaseto: It was intentional since 1.11's primary dependency management is still vendor so I figured it would be best to test for both until 1.13 comes out and defaults to modules. If you'd like to only test for 1.11 modules that's totally cool with me though and I can move tip down there as well.

Makefile
zaquestion: The problem is, the below deps are not merely imports used in truss itself, but binaries that truss calls out to. The goal of `make dependencies` is to get those executables in your PATH. I don't believe `go mod download` nor enforcing them in the `go.mod` will function. The CI is a good test if this however, so if you see it working there you're probably in the clear.

kainoaseto: Ah I see, yeah I think you are right. That is kind of frustrating that they won't do installs from modules as well as downloading them considering it's meant to be the main dependency management option going forward. Hm might still be a place for custom tools to wrap modules and handle this all properly.

zaquestion (approved):

PR #268, opened 2019-05-13
Allow customization of client CLI output.

TyeMcQueen:
Here is a fairly small change that allows custom code to be added to control what the CLI client outputs.  The default code behaves very similar to how things used to work but also includes a commented-out example of how to make the client output JSON instead.

I personally would be in favor of making JSON output the default, but I wanted to minimize the changes.  If JSON output as default is desired, I'd be happy to submit another pull request that does that after this one is applied.


zaquestion:
Honestly the cli clients are fair game. Other than a desire to polish them as cobra clients ( https://github.com/metaverse/truss/issues/232 ) there isn't a lot planned for them. To my knowledge folks have only used them for debugging, but mostly use `curl` or `httpie` to interact with them. Default JSON makes enough sense to me, I'd leave the other version in as a `-v`, `--verbose` option. If you put it up I'll get a few folks in the community to thumb it up.

TyeMcQueen:
I am suspicious that the travis failures are transient but I have yet to find how to request that they be re-run (other than adding some cosmetic commit).

zaquestion:
I'll give it a rerun, apologies if ive just been overlooking them broken. They have integration test that test the real stuff and some of that is constantly changing under our feat as we try to stay current with certain dependencies

zaquestion:
I'll take a closer look this week and merge if it's unrelated. Will also try to fix the tests of course

zaquestion:
```
```
Will be addressed in https://github.com/metaverse/truss/pull/274, but aren't related the this changes in this PR. Reached out to #truss on the gophers slack to get a few thumbs up.

zaquestion:
@TyeMcQueen Can merge/rebase in the latest changes and that should resolve the tests.

TyeMcQueen:
Actually, can't since "From those we have talked to about truss, we have never met anyone whose actually used them".

zaquestion:
Hey @TyeMcQueen, apologies I should have updated here after we made that decision. We considered your usage when removing them, it was incorrect to say we haven't met anyone. What follows from that quote in https://github.com/metaverse/truss/pull/282
> Short of some recent PRs to expand some CLI functionality they've remained untouched, we noticed the submitters of these recent changes to be off on there own fork and likely to be unaffected by these changes.

Definitely open to a dialog here, and particularly interested in how you're using the CLIs and the value they've brought you. Since removing the CLIs I took on an ambitious project to support `oneof` in the query, it was intended to be experimental, but we think it turned out well, so it's entirely possible we _could_ actually support `oneof` for CLIs (and prevent truss services from failing complication when used).

If ultimately getting back on the metaverse upstream is a goal, I certainly don't want this to be a blocker. With some dialog I'm more than happy to entertain bringing the CLIs back or even putting in the effort to polish them as cobra CLIs. Would love to hear back from you here or over in the #truss channel in the gophers slack. 

Thank you for using truss, it's truly an honor to see it out in the wild.

TyeMcQueen:
Thanks for the kind and reasonable response!

My usage of the truss CLI client is very similar to how I use the AWS CLI and GCP CLI (gcloud) and the features I have added and plan to add line up with features in both of those that I use frequently.

I've already changed the CLI client to output JSON which makes it as useful as curl.  (Before my minor improvements, I'm not surprised that nobody used the CLI client because the output format bordered on useless.) And I've implemented a few additional improvements but have more planned.  

Currently, for me, using curl instead of the CLI means:

1) Everything has to be entered as JSON, which requires lots more punctuation.
2) There are no usage messages to help jog your memory when making a request.
3) Having to enter the complex address to be reached rather than just having that defaulted.
4) Having to know which HTTP method with which path rather than just knowing the name of the operation I want to invoke.
5) Having to specify custom headers.

"foo listrecords --limit 10" vs.
"curl -X POST -H ... -d '{"limit": 10}' https://mysvc.myteam.internal.company.com/api/v1/record"

I plan to enhance the generated CLI clients to be much more customizable such as allowing data structures to be passed as YAML so tons of quotes can be skipped (I don't mind the hidden complexity of YAML's implementation in an optional client but do not want to add that complexity to the service where it greatly widens attack surfaces).

Similarly, the client can allow much more human-friendly formats for entering dates (also something I don't want in the service since complex parsing is a very common source of security vulnerabilities).  For example, we have a shorthand for dates of "MMDD" because of some tools we use so I have CLI clients outside of truss that support "1101..1201" as short for "--since 2019-11-01T00:00 --until 2019-12-01T00:00".  It won't be hard to add the ability to specify custom hooks in the truss CLI client that would allow preprocessing of arguments to support such labor-saving shortcuts.

Next will be adding auth support via customizable hooks which will be a huge win over what auth will require when using curl.

The ability to output YAML and/or flattened data structures will likely also happen.

Also, FWIW, I avoid the use of 'oneof' because of how it doesn't quite "fit in".  So I still plan to use the CLI client even if it comes with a "does not support 'oneof'" caveat.

I hope you see the potential value of such clients so my work can be part of truss, not just in a fork of the project.

zaquestion:
Thanks for the detailed response, those usecases make sense to me, and I wouldn't have guessed that you had so much intention to improve them when making the decision to rip them out. 

>  (Before my minor improvements, I'm not surprised that nobody used the CLI client because the output format bordered on useless.)

well said

The main problem with the clis and `oneof` was that it would break compilation if `oneof` was used. IIRC that included building the server because of some coupling, but I could be misremembering that, and that could be worked around. 

We'd still like `cmd/NAME` to refer to the server (formerly `cmd/NAME-server`) instead of the cli going forward and I don't quite have the time right now to reincorporate the cli clients. If you'd like to take on bringing them back in I think we can work them back in. Ideally they're decoupled enough that it only takes deleting 2 folders (1 in `svc` and 1 in `cmd`) to remove them from a generated project. In the long run I think I can get `oneof` actually working with them, but so long as it doesn't break compiling the server I see no reason we can't bring them back atm. If you go down this path, perhaps create a separate issue to discuss further (if needed) or we can discuss when a PR comes up.

For reference the changes to support `oneof` in the query came in here: https://github.com/metaverse/truss/pull/284

PR #269, opened 2019-05-13
Sort CLI flags returned by AllFlags() so generated code is stable.

TyeMcQueen:
This is small change to add minimal sorting to AllFlags() so regenerating the code with no changes to the protobuf def'n does not generate differences due to random sorting of the lines declaring the CLI flag variables.  Also reduces the number of changes generated when small changes are made to the protobuf def'n.


zaquestion (approved):

zaquestion:
Thanks, this will be a nice QoL improvement :)

PR #274, opened 2019-06-08
update to latest gogo/protobuf, only test with modules enabled

zaquestion:


lelandbatey (approved):
I think these changes move us forward even if they don't fix everything. For that reason, I approve.

PR #277, opened 2019-07-18
Remove testing on tip in favor of testing only stable releases

zaquestion:


lelandbatey (approved):
Since tip causes more problems than it solves, I think removing testing against tip is the thing to do.

PR #278, opened 2019-07-18
use gogo's jsonpb

zaquestion:
Changes in https://github.com/metaverse/truss/pull/227 brought in jsonpb, but only occured to me after merge that they used the `golang/protobuf` package, but we switched to `gogo/protobuf` back in https://github.com/metaverse/truss/pull/262/commits/249f172d85815fa8449b3b90d74cb7d380a57204, so these changes are just getting everything on the same page. That said, I have no idea what the implications of not merging this are.


lelandbatey (approved):
Good catch, LGTM!

PR #282, opened 2019-08-01
drop support for generated CLI clients

zaquestion:
Truss generated CLI clients have been a long unmaintained "feature", now that truss supports features like `oneof` and the use of jsonpb marshalers, its become harder to keep the clients working for all proto options. `oneof` specifically would cause CLI clients to be generated in unbuildable states, and we didnt have a way to solve this. Additionally we believe this to be a severely underutilized feature. From those we have talked to about truss, we have never met anyone whose actually used them, we have never used them in our own truss projects. Short of some recent PRs to expand some CLI functionality theyve remained untouched, we noticed the submitters of these recent changes to be off on there own fork and likely to be unaffected by these changes.

CLI generation as a redeveloped feature would still be considered, but in their current state its time to drop the axe


zaquestion:
Could use advise on how we should notify users when running truss about the changes. And would like more dialog on if/how we should cleanup stale (potentially build breaking) cli files.

lelandbatey:
:fire: BURN IT ALL DOWN :fire:

lelandbatey (dismissed):

lelandbatey (approved):

PR #283, opened 2019-08-24
obviously the enums should be strings

zaquestion:
this got overlooked from the orginal jsonpb behavior since submitter was trying to maintain behavior, but strings is way better

before
```
$ curl -X POST localhost:5050/rack -d { "env": "prod" }
{"env":1"}
```

now
```
$ curl -X POST localhost:5050/rack -d { "env": "prod" }
{"env":"prod"}
```


mauricebutts (approved):
LGTM

PR #284, opened 2019-08-25
support oneof parameters in query

zaquestion:


gengokit/httptransport/templates/client.go
```diff
@@ -217,15 +238,15 @@ func contextValuesToHttpHeaders(keys []string) httptransport.RequestFunc {
 	// error and attempt to decode the specific error message from the response
 	// body. Primarily useful in a client.
 	func DecodeHTTP{{$method.Name}}Response(_ context.Context, r *http.Response) (interface{}, error) {
+		defer r.Body.Close()
```
zaquestion, line 213: @lelandbatey Am I missing something or were we really not doing this??

gengokit/httptransport/templates/server.go
```diff
@@ -8,6 +8,7 @@ var ServerDecodeTemplate = `
 	// decodes a JSON-encoded {{ToLower $binding.Parent.Name}} request from the HTTP request
 	// body. Primarily useful in a server.
 	func DecodeHTTP{{$binding.Label}}Request(_ context.Context, r *http.Request) (interface{}, error) {
+		defer r.Body.Close()
```
zaquestion, line 11: @lelandbatey Am I missing something or were we really not doing this??

gengokit/httptransport/templates/client.go
lelandbatey: We really weren't doing this. I think the reason we weren't doing this is because of [this line from the documentation on `http.Request.Body`:](https://golang.org/pkg/net/http/#Request.Body)

```
...
The Server will close the request body. The ServeHTTP
Handler does not need to.
```
I'm too under the weather at this moment to say for sure from my reading of that comment whether that means we definitely do not need to close `request.Body` or not, but I don't think adding these `defer r.Body.Close()` statements will hurt anything.

zaquestion:
@lelandbatey bringing this in since it now contains efforts to fix `go1.13` issues. Realistically could split these up, but given our inperson review of the oneof in query logic, I'm opting to bring it in.

PR #288, opened 2019-10-13
Add paths=source_relative option to restore placement of output files.

TyeMcQueen:
I'll admit that I'm a little confused, but I _think_ that an update to an underlying dependency changed the default behavior about into which directory generated `*.pb.go` files get written.

This small change made my automation start working as it used to, even though I failed at finding documentation about this apparent change in behavior.

(If anybody has an idea what might have caused this change where generated `*.pb.go` files get written into a deep subdirectory of the specified output directory, a pointer to it would be appreciated.)

### DIFF
--- a/truss/execprotoc/execprotoc.go
+++ b/truss/execprotoc/execprotoc.go
@@ -23,7 +23,7 @@ func GeneratePBDotGo(protoPaths, gopath []string, outDir string) error {
 		"Mgoogle/protobuf/struct.proto=github.com/gogo/protobuf/types," +
 		"Mgoogle/protobuf/timestamp.proto=github.com/gogo/protobuf/types," +
 		"Mgoogle/protobuf/wrappers.proto=github.com/gogo/protobuf/types," +
-		"plugins=grpc:" + outDir
+		"paths=source_relative,plugins=grpc:" + outDir
 
 	_, err := exec.LookPath("protoc-gen-gogo")
 	if err != nil {


TyeMcQueen:
It may have been moving to go 1.13 ?  I didn't notice a team mate had done that in our project.

zaquestion:
@TyeMcQueen I've noticed some fucky behavior with `go1.13` and I've been completely lost on how to fix. I'll try this out with the branch I was working on and possibly be forever in your debt because I've been pretty lost on a fix. Thank you!

zaquestion:
@TyeMcQueen Can you share anymore information on what you ran into. I've been seeing issues with our tests and go1.13, but they aren't resolved with this fix unfortunately. I also tried generating a new service with `go1.13` and see a similar issue as in the tests.

That issue being that modules have stopped being correctly reference and imports are coming as absolute paths, like so -- this happens regardless of the changes in this MR
```
 pb "/home/zaq/projects/trussstest"
```

It doesn't happen on `go1.12` however. My protoc version is:
```
$ protoc --version
libprotoc 3.6.1
```
I believe I've seen it work on `3.7.x` at least as well.

Happy to merge this in, just hoping to improve my own understanding of the changes first and do proper diligence

zaquestion:
Did you start using `go_package` in your proto files by chance? It seems like that `source_relative` option is being used as a response to behavior cause by that flag.

https://github.com/golang/protobuf/issues/515

zaquestion:
Hmmm, well I fixed the weird incorrect module path issues I was seeing by wiping out truss's `go.mod` and recreating from scratch in `go1.13`, so if you should see that it'll be fixed when https://github.com/metaverse/truss/pull/284 comes in (ideally merged by the time you see this, waiting on tests)

TyeMcQueen:
Yes, we are using `go_package`.  We had been using that before I had to make this patch so it sounds like an update to something added this option to fix a problem with using `go_package` that was not bothering us.

So this merge request should help others who use `go_package` with up-to-date dependencies.

zaquestion:
@TyeMcQueen Thanks for the added context, appreciate it!

PR #292, opened 2019-11-27
support custom server options when constructing handlers

zaquestion:
useful when implemention your own `main.go`


gengokit/template/NAME-service/svc/transport_grpc.gotemplate
```diff
@@ -21,10 +21,11 @@ import (
 )
 
 // MakeGRPCServer makes a set of endpoints available as a gRPC {{.Service.Name}}Server.
-func MakeGRPCServer(endpoints Endpoints) pb.{{.Service.Name}}Server {
+func MakeGRPCServer(endpoints Endpoints, options ...grpctransport.ServerOption) pb.{{.Service.Name}}Server {
 	serverOptions := []grpctransport.ServerOption{
 		 grpctransport.ServerBefore(metadataToContext),
 	}
+        serverOptions = append(serverOptions, options...)
```
adamryman: Whitespace :wink:

adamryman (dismissed):
Pending the small nit of whitespace. Looks good to me.

As they are variadic, this should not cause issues for existing users in most cases.

gengokit/template/NAME-service/svc/transport_grpc.gotemplate
zaquestion: grr, no `gofmt` in templates, good catch.

adamryman (approved):
:shipit:

toctan:
@zaquestion I was wondering if it's possible to make it a little bit easier to provide custom server options without writing a new `main.go`, since a lot of packages in `go-kit` require some of these server options, for example, jwt and zipkin tracing. Can we provide a function similar to `handlers.WrapEndpoints` and `handlers.WrapService` so that the user could easily inject custom server options into `svc/server/run.go`? Maybe something like `handlers.CustomServerOptions`, or we could put it into another package like transport? This should be pretty easy to implement. I could send out a pull request if you think this is a good idea, thanks!

zaquestion:
Generally we recommend creating your own `main.go` since it's pretty straightforward and expands the control available to users who want more custom options. Historically, we haven't seen this as a pain point, and it helps keep "default truss" as transport agnostic as possible. I would certainly entertain a PR or discussion in a new Issue, one thing I'll say here is that there are custom server options that can be passed to grpc and/or http/json handlers, so we'd likely want ways to configure them individually, which could get messy down the line as additional transports are added.

Ref: https://github.com/metaverse/truss/wiki/Using-a-custom-main.go-with-generated-truss-services

For what it's worth, I use custom `main.go` files in many of my personal/work projects, particularly when I want to use truss http only.

toctan:
Got it. I didn't know it was recommended to create my own `main.go`, but it definitely works for my use case. Thanks!

I also realized that we need to have different server options for different transports, so I was thinking something like this:
```go
// InjectGRPCServerOptions provides a list of server options to grpctransport.NewServer.
func InjectGRPCServerOptions() []grpctransport.ServerOption {
	options := []grpctransport.ServerOption{
		grpctransport.ServerBefore(jwt.GRPCToContext()),
	}
	return options
}

// InjectHTTPServerOptions provides a list of server options to httptransport.NewServer.
func InjectHTTPServerOptions() []httptransport.ServerOption {
	options := []httptransport.ServerOption{
		httptransport.ServerBefore(jwt.HTTPToContext()),
	}
	return options
}
```

PR #299, opened 2020-07-24
fix: parsing errors when using foreign types in rpcs

zwiedmann-isp:
This doesn't add support for foreign types, but does allow truss
generation to complete without errors. Why?? This allows for developers
to add type aliases alongside the pb.go file generated by `protoc` which
reference the desired types.

```
  rpc GetAccount(Empty) returns (google.protobuf.Struct) {
    option (google.api.http) = {
      get: "/accounts/me"
    };
  }
```
will still generate a handler that looks like
```
func (s recurlyService) GetAccount(ctx context.Context, in *pb.Empty) (*pb.Struct, error) {
```
Note: pb.Struct as the (technically) incorrect return type.

However, since code generation succeeded, only 1 additional line is
needed for this to function correctly, and it can be added outside
generated code space

```
type Struct = types.Struct
```
placed in a separate file in the same package as one's pb.go this allows
the protobuf types to be referenced without a wrapper message. This is
primary useful for the "WellKnownTypes" (see ref), which have special
handling in protobuf, particularly in jsonpb and can't be replicated
simply by copying the message into the service proto. For non-special
external types, it may make more sense to simply copy the message
instead of using the technique allowed here

Ref: https://developers.google.com/protocol-buffers/docs/reference/google.protobuf


lelandbatey (approved):

HasAlNguyen:
Hey guys - how do I get off the distribution list for all the notifications?!

zwiedmann-isp:
@HasAlNguyen Nice to see your name popup :D 

You just need to unwatch the repository, see the dropdown towards the upper right of the webpage. Alternative, there might be an unsubscribe at the bottom of the emails.

PR #301, opened 2020-08-11
decouple NewService from NewEndpoints & govet fixes

zwiedmann-isp:
This change improves the flexibility of interacting with the gokit
endpoints and service particularly when creating one's own `main.go`,
thus improving [Support Level 2](https://github.com/metaverse/truss/wiki/Levels-of-support-for-different-truss-usecases)

The specific usecase prompting this change is to allow adding an
additional gokit endpoint to the generated gokit service outside truss
generation. In our case because we need to support an xml handler in our
service


zaquestion (approved):

lelandbatey (commented):
This kind of general architectural cleanup is super welcome, thank you for these changes!

PR #302, opened 2020-08-13
perf: skip walking unnecessary dirs when reading previous generation

zaquestion:
Noticed an error like this after running truss with a flattened service
path.
```
truss --svcout . service.proto
FATA[0000] cannot parse input: cannot read previously generated files: cannot fully walk directory /home/zaq/projects/go-mod/src/github.com/<redacted>: cannot read file: /home/zaq/projects/go-mod/src/github.com/<redacted>/.git/refs/remotes/origin/<redacted>: open /home/zaq/projects/go-mod/src/github.com/<redacted>/.git/refs/remotes/origin/<some_branch>: too many open files
```
NOTE: was using default nofiles limit on linux
```
$ ulimit -Sn
1024
```
When using `--svcout .`, `cmd`, `svc`, and `handlers` folder will
be created in the current directory. Which is really nice if you want a
flatter service structure. However, this has the unintented consequence
of opening significantly more files when truss is building context on
it's previous generation. This change cuts down the files read to only
those in the handlers dir as these are the only files whose previous
generation is leveraged

Ref: `gengokit/generator/gen.go:generateResponseFile`

### DIFF
--- a/cmd/truss/main.go
+++ b/cmd/truss/main.go
@@ -359,15 +359,23 @@ func readPreviousGeneration(serviceDir string) (map[string]io.Reader, error) {
 		return nil, nil
 	}
 
+	const handlersDirName = "handlers"
 	files := make(map[string]io.Reader)
 
 	addFileToFiles := func(path string, info os.FileInfo, err error) error {
 		if info.IsDir() {
-			return nil
+			switch info.Name() {
+			// Only files within the handlers dir are used to
+			// support regeneration.
+			// See `gengokit/generator/gen.go:generateResponseFile`
+			case filepath.Base(serviceDir), handlersDirName:
+				return nil
+			default:
+				return filepath.SkipDir
+			}
 		}
 
 		file, ioErr := os.Open(path)
-
 		if ioErr != nil {
 			return errors.Wrapf(ioErr, "cannot read file: %v", path)
 		}


cmd/truss/main.go
```diff
@@ -363,11 +363,18 @@ func readPreviousGeneration(serviceDir string) (map[string]io.Reader, error) {
 
 	addFileToFiles := func(path string, info os.FileInfo, err error) error {
 		if info.IsDir() {
-			return nil
+			switch info.Name() {
+			// Only files within the handlers dir are used to
+			// support regeneration.
+			// See `gengokit/generator/gen.go:generateResponseFile`
+			case filepath.Base(serviceDir), "handlers":
```
adamryman: Consider extracting `"handlers"` as a constant at the top of this function.

I wish things were organized in such a way that this directory name was exposed, though such is life.

adamryman (dismissed):
This is a great catch!

One comment, though I think what you have is fine.

If there is ever a major refactor, this would be a good thing to fix.

LGTM.

adamryman (approved):
Ok, I thought for some reason that `hooks.go` and `middlewares.go` were in a directory other than the `handlers` directory.

I suppose a name other than `handlers` would have been more clear, though such is how things go as software changes.

Extracting the `handlers` string into a `const` would be preferred. Though this looks good to merge either way.

:+1:

cmd/truss/main.go
zaquestion: Extracted into a const

PR #306, opened 2020-08-26
Support using protobuf types in query parameters

zaquestion:
Useful for passing dates in the query

### DIFF
--- a/gengokit/httptransport/httptransport.go
+++ b/gengokit/httptransport/httptransport.go
@@ -11,9 +11,9 @@ import (
 	"text/template"
 	"unicode"
 
-	log "github.com/sirupsen/logrus"
 	gogen "github.com/gogo/protobuf/protoc-gen-gogo/generator"
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 
 	"github.com/metaverse/truss/gengokit/httptransport/templates"
 	"github.com/metaverse/truss/svcdef"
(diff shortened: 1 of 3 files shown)


zaquestion:
Honestly I think a lot of this json unmarshaling in the query stuff is going to be up for the chopping block in `v1`

zaquestion:
This need to wait for 

- [ ] https://github.com/metaverse/truss/pull/304 (make corresponding changes here too)
- [ ] Tests

gengokit/httptransport/httptransport.go
```diff
@@ -381,6 +399,10 @@ func createDecodeConvertFunc(f Field) (string, bool) {
 var {{.LocalName}} *{{.GoType}}
 {{.LocalName}} = &{{.GoType}}{}
 err = json.Unmarshal([]byte({{.LocalName}}Str), {{.LocalName}})`
+		singlePBTypeUnmarshalTmpl := `
+var {{.LocalName}} *{{.GoType}}
+{{.LocalName}} = &{{.GoType}}{}
+err = jsonpb.UnmarshalString({{.LocalName}}Str, {{.LocalName}})`
```
toctan, line 405: Instead of creating thses tailored decode functions for each individual field, how about we we first merge all the variables from path, query, and body into a single JSON string, then call `jsonpb.Unmarshal(s, req)` to decode into the `request` struct in one go? Seems like it would simplify the decode logic significantly?

toctan:
This is great! We need this to use the `FieldMask update_mask` field in the standard update method: https://cloud.google.com/apis/design/standard_methods#update

kmk142789 (approved):

PR #310, opened 2020-08-31
(ci) use goreleaser to manage releases

zaquestion:
Will build truss for Mac (+push to homebrew tap) and linux. Will add binaries and a generated changelog to the github release


adamryman (approved):
LGTM

PR #320, opened 2020-12-19
Remove --pbout from readme

avonar:
it seems pbout does not exist anymore.
 https://github.com/metaverse/truss/commit/d1119d3a68ab1f41e850ef18930504c0c43b2739


zaquestion:
Good catch, and correct, `pb.go` files always generate alongside their `.proto` files now.

PR #326, opened 2021-02-08
Mlj

skyfile:


.gitignore
```diff
@@ -15,3 +15,5 @@ _example/*
 # Prevents accidentally including a problematic folder within the `vendor/`
 # folder.
 vendor/github.com/sirupsen/logrus/examples/
+go.mod
+go.sum
```
zaquestion, line 19: you actually want to commit these

cmd/truss/main.go
```diff
@@ -175,10 +181,11 @@ func parseInput() (*truss.Config, error) {
 
 	svcName = strings.ToLower(svcName)
 
-	svcDirName := svcName + "-service"
+	//svcDirName := svcName + "-service"
+	svcDirName := ""
```
zaquestion, line 185: if you're trying to get the service to generate in the current directory, no codechange is needed. you can pass `--svc-out=./`

PR #330, opened 2021-06-04
Allow hooks.go to control server configuration e.g. EmitDefaults

lelandbatey:
This change adds a `SetConfig()` function to `<svcname>/handlers/hooks.go` which allows user-modifiable code to make changes to the configuration used to start up a truss service.

To leverage this new configuration power, a new configuration parameter has been added to the `server.Config` struct. This change adds a `GenericHTTPResponseEncoder` field to the `server.Config` struct, a function which replaces the previously hard-coded `EncodeHTTPGenericResponse`. This allows users total control over how they're encoding the responses of their HTTP requests, including the possiblity of setting `EmitDefaults` to be `true` (which is the use-case at the origin of this change.)


zaquestion (approved):
Excellent! thank you for expanding on the comments too!

adamryman (approved):
Looks great!

Comments are excellent.

PR #334, opened 2021-07-14
fix bug: package "io" is not used

kvii:
package "io" is not used when you not use http annotations in .proto file


zaquestion:
Ah, this does look like an oversight, good catch! Sorry for the inconvenience

zaquestion (approved):

zaquestion:
Released in `v0.3.1`

PR #339, opened 2021-09-11
fixes warning 'installing executables with 'go get' in module mode is…

vpiyush:
… deprecated.'


zaquestion:
guessing thag you noticed that this isnt deprecated, however, as of go 1.16 theres a slightly better way to do this using `go install` might actually be a good idea to switch these over to that

